PR #3, opened 2015-07-08
Debug electron ID

OlivierBondu:
Now with non-trivial electron ID maps

there is still some weird things in there: for example in some cases the mva tight ID passes when the mva loose ID does not...


blinkseb:
Wo, thanks for this one!

PR #14, opened 2015-07-27
Adding MC matching info for jets and leptons

OlivierBondu:


OlivierBondu:
@blinkseb : any comment is welcome :smile:

blinkseb:
I guess we can avoid code duplication by adding an additional argument to the `fill_candidate` method, like

````C++
template <typename T, typename U>
void fill_candidate(const T& p, const U* gen) {
}
```
And calling it with the right function call from the producers

OlivierBondu:
mmm, yes, but what to do for other types depending on CandidatesProducer that don't have gen info (HLT for example?)

blinkseb:
Passing ``nullptr`` should be enough?

OlivierBondu:
refactored

PR #21, opened 2015-08-04
Improvement of categories

blinkseb:
@AlexandreMertens made some wise suggestions concerning categories management. This is an attempt to fix the situation.

Currently, the path followed by the code is:
 - Run producers
 - Run analyzers
 - Run categories
 - If an event belongs to no category, then drop it, otherwise, save everything

The issue here is you can't access categories information from the analyzer, because the code is ran after. An easy solution would be to invert the categories and analyzers step, but now you cannot longer access any data produced by the analyzers.

I opted for the following solution:
 - Run producers
 - Run categories pre-analyzers
 - If the event both now belongs into any categories, drop it and pass to the next event
 - Else, run analyzers
 - Run categories post-analyzers
 - If an event belongs to no category, then drop it, otherwise, save everything

A category now has two callbacks, ``bool event_in_category_pre_analyzers(const ProducersManager&)`` and ``bool event_in_category_post_analyzers(const ProducersManager&, const AnalyzersManager&)``. Now the tricky part:

 - For an event to belong to a category, both function has to return ``true``
 - ``bool event_in_category_pre_analyzers(const ProducersManager&)`` is called first, before the analyzers
 - ``bool event_in_category_post_analyzers(const ProducersManager&, const AnalyzersManager&)`` is only called if the pre-analyzers callback returns ``true``

Also, the ``evaluate_cuts`` is now splitted in ``evaluate_cuts_pre_analyers`` and ``evaluate_cuts_post_analyzers``.

If think this solution provides the flexibility we need. It still remains to add a way to get categories and cuts info from the analyzers, but it should be easy to do.

What do you think?

**Note**: This depends on #20 so it must be merged after. Please ignore first commit.

### DIFF
--- a/interface/Analyzer.h
+++ b/interface/Analyzer.h
@@ -10,7 +10,6 @@
 #include <FWCore/Framework/interface/ConsumesCollector.h>
 #include <FWCore/Utilities/interface/InputTag.h>
 
-#include <cp3_llbb/Framework/interface/Category.h>
 #include <cp3_llbb/Framework/interface/ProducersManager.h>
 #include <cp3_llbb/TreeWrapper/interface/TreeWrapper.h>
 
@@ -20,6 +19,8 @@
 #include <map>
 
 typedef ROOT::Math::LorentzVector<ROOT::Math::PtEtaPhiE4D<float>> LorentzVector;
+
+class CategoryManager;
 
 namespace Framework {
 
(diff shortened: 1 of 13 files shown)


AlexandreMertens:
Great!
Yes, I also think that this solution provides the needed flexibility. Thanks!

AlexandreMertens:
Hello Seb,

How is it possible to store the category info into the rootfile?
Isn't the category post analyser supposed to do it?

OlivierBondu:
It should be already the case ?

in the test example the category info is stored as 'two_muons_category', 'two_muons_muon_1_pt_cut' and such, you do not have these present in the output tree ?

PR #22, opened 2015-08-05
Store sum of event weight

blinkseb:
This PR adds a basic metadata manager allowing a producer to store metadata into the files. For now, only float and double are supported, but it's easy to extend it on demand for other (more complex) types.

This metadata manager is used to store the sum of event weight into the final file. This sum is also printed on stdout in case something needs to parse it.


OlivierBondu:
perfect, many thanks :)

PR #23, opened 2015-08-17
Making the btag discriminators accessible from the producers

swertz:
Add a variable to JetsProducer to make to btag discriminatros accessible from the producer.

(this is more of a suggestion open for discussion... I'd be happy to lear if there's a better way to do this. I've tried to avoid creating the branches in the loop, but rather in the constructor of the JetProducer, from the edm::parameter set passed to it, but that does't seem possible)


blinkseb:
Indeed, currently it's not possible to access b-tag discriminators from the analyzers... We discussed about that a bit today with @OlivierBondu, @AlexandreMertens and @vidalm and it's been decided that, since we need to know the btag discriminator ugly name to access it from the analyzer, we'll only stores discriminators requested by the users in the python configuration file. Then, these discr will be available through a map to the analyzers.

I'll try to open a new PR with these changes today or tomorrow.

OlivierBondu:
@swertz : can we close this one ?

Issue #25, opened 2015-08-28
Access to eID from the analyzer


OlivierBondu:
Given the Identifiable::ids is protected, one cannot access them through 
```C++
electron.ids[ielectron]["mycomplicatedelectronIDname"]
```
see: https://github.com/cp3-llbb/Framework/blob/master/interface/Identifiable.h#L38

shouldn't it be moved in public instead ?

I just want to make sure I am not breaking anything because on the other hand the ids are written to the tree, and the tree members are (already) supposed to be public: 
https://github.com/cp3-llbb/Framework/blob/master/interface/ElectronsProducer.h#L31-L32

PR #26, opened 2015-08-31
Moving Identifiable::ids from protected to public

OlivierBondu:
Fix https://github.com/cp3-llbb/Framework/issues/25


blinkseb:
Thanks :)

PR #27, opened 2015-08-31
Make btagging discriminants accessible from analyzers

blinkseb:
This is an attempt to fix the issue raised by #23. Only b-tagging discriminants requested by the user in the python configuration are considered. A branch per discriminant is still created, and one can access the b-tagging discriminant by using the function

```C++
float JetProducer::getBTagDiscriminant(size_t index, const std::string& name);
```

### DIFF
--- a/src/JetsProducer.cc
+++ b/src/JetsProducer.cc
@@ -22,16 +22,14 @@ void JetsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup)
 
         vtxMass.push_back(jet.userFloat("vtxMass"));
 
-        auto& btag_discris = jet.getPairDiscri();
-        for (auto& btag_discri: btag_discris) {
-            // Get branch from tree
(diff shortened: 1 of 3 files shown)


src/JetsProducer.cc
```diff
@@ -22,16 +22,14 @@ void JetsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup)
 
         vtxMass.push_back(jet.userFloat("vtxMass"));
 
-        auto& btag_discris = jet.getPairDiscri();
-        for (auto& btag_discri: btag_discris) {
-            // Get branch from tree
-            std::vector<float>& branch = tree[btag_discri.first].write<std::vector<float>>();
-            branch.push_back(btag_discri.second);
+        for (auto it: m_btag_discriminators) {
```
swertz: Shouldn't it be auto &it?

blinkseb:
Updated as per comments

PR #32, opened 2015-09-07
Refactor Framework python configuration

blinkseb:
This PR intent to fix the current mess with analyzer's configuration file.

A ``create`` function is introduced. Just call this function from any configuration file to create the CMSSW ``process`` and configure the framework. The ``create`` function has doctype attached which describes how-to use it, so it should be really easy to use.

Two new configuration files are added, one for MC and one for data, illustrating how to use the new function.


test/TestConfigurationData.py
```diff
@@ -0,0 +1,17 @@
+
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.StandardSequences.Eras import eras
```
swertz: Hi, is that a CMS thing or is a file missing from the commit?

blinkseb: It's a CMS thing :
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCmsDriverEras

2015-09-07 13:07 GMT+02:00 Sébastien Wertz <notifications@github.com>:

> In test/TestConfigurationData.py
> <https://github.com/cp3-llbb/Framework/pull/32#discussion_r38853486>:
>
> > @@ -0,0 +1,17 @@
> > +
> > +import FWCore.ParameterSet.Config as cms
> > +
> > +from Configuration.StandardSequences.Eras import eras
>
> Hi, is that a CMS thing or is a file missing from the commit?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cp3-llbb/Framework/pull/32/files#r38853486>.
>



-- 
Sébastien Brochet

swertz: K, thanks!

PR #33, opened 2015-09-08
Adding dilepton categories / analyzer

OlivierBondu:
In my understanding this should be enough to run the dilepton_analyzer before the test_analyzer, but I am not sure what would need to be duplicated in the {TT,HH}Analysis packages to be sure it is run ?

~~I have one more commit I'd like to add before merging (renaming dimuons into dileptons_mumu, etc. for overall consistency)~~
Now good to go on my end

### DIFF
--- a/plugins/DileptonCategories.cc
+++ b/plugins/DileptonCategories.cc
@@ -30,8 +30,12 @@ void MuMuCategory::evaluate_cuts_pre_analyzers(CutManager& manager, const Produc
 
 void MuMuCategory::evaluate_cuts_post_analyzers(CutManager& manager, const ProducersManager& producers, const AnalyzersManager& analyzers) const {
     const DileptonAnalyzer& dilepton_analyzer = analyzers.get<DileptonAnalyzer>("dilepton_analyzer");
-    if( dilepton_analyzer.dileptons_mumu[0].M() > 20. )
-        manager.pass_cut("ll_mass");
+    for(unsigned int idilepton = 0; idilepton < dilepton_analyzer.dileptons_mumu.size(); idilepton++)
+        if( dilepton_analyzer.dileptons_mumu[idilepton].M() > 20. )
+        {
+            manager.pass_cut("ll_mass");
+            break;
+        }
 }
 
 // ***** ***** *****
@@ -60,8 +64,12 @@ void MuElCategory::evaluate_cuts_pre_analyzers(CutManager& manager, const Produc
 
 void MuElCategory::evaluate_cuts_post_analyzers(CutManager& manager, const ProducersManager& producers, const AnalyzersManager& analyzers) const {
     const DileptonAnalyzer& dilepton_analyzer = analyzers.get<DileptonAnalyzer>("dilepton_analyzer");
-    if( dilepton_analyzer.dileptons_muel[0].M() > 20. )
-        manager.pass_cut("ll_mass");
+    for(unsigned int idilepton = 0; idilepton < dilepton_analyzer.dileptons_muel.size(); idilepton++)
+        if( dilepton_analyzer.dileptons_muel[idilepton].M() > 20. )
+        {
+            manager.pass_cut("ll_mass");
+            break;
+        }
 }
 
 // ***** ***** *****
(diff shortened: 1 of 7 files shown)


plugins/DileptonCategories.cc
```diff
@@ -0,0 +1,126 @@
...
+    return dilepton_analyzer.dimuons.size() > 0;
+};
+
+void MuMuCategory::register_cuts(CutManager& manager) {
+    manager.new_cut("ll_mass", "mll > 20");
+};
+
+void MuMuCategory::evaluate_cuts_pre_analyzers(CutManager& manager, const ProducersManager& producers) const {
+    const MuonsProducer& muons = producers.get<MuonsProducer>("muons");
+    if( (muons.p4[0] + muons.p4[1]).M() > 20.)
```
swertz: Hi Olivier,
I don't understand this cut. One could imagine the cut being false, but while having the first(second) muon not passing muonID, and the second(first) and third muons passing muonID and having m_ll > 20. I this case we reject the event but we do have a good dimuon pair. 
Is there are reason for not keeping those events which I'm not seeing?

blinkseb:
Assigning a 'Do not merge' label until I test and push some things I already discussed with Olivier.

plugins/DileptonCategories.cc
OlivierBondu: very fair point, which I noticed then forgot... correcting this

OlivierBondu:
@swertz : 
  * I removed the evaluate_cut_pre_analyzer as this would require looping over the leptons... hence redoing the combinatorics already being performed in the analyzer itself, it seems unecessary duplication at this point.
  * I modified the evaluate_cut_post_analyzer to check that at least one pair (not necessarily the first one) pass the invariant mass cut

thanks for spotting !

swertz:
Thanks for the update! I'm unable to run any tests right now so I'll let @blinkseb take care of the merge...

blinkseb:
I'll test more in to depth tomorrow. I think we need to remove hardcoded things, like the electron id name. These must be a configured via the python file. Maybe also an option to specify is we want to use loose, medium or tight ID for the leptons? But this can be added in a future iteration.

Ideally, the value of the mll cut should also be customizable, but it's not easily possible for the moment I'm afraid.

OlivierBondu:
Yes, I agree that in principle we want to be able to specify which level of ID we want, but I think (for DoubleLepton datasets) that loose lepton ID is the minimum we want, even for possible control samples ? mll > 20 GeV in any case is supposed to be loose for all our signals

for customizing the mll cut, can't we just specify the value itself, not the variable, as a double in some python config that would be specific to this analyzer  (not generic to any analyzer) ?

blinkseb:
Yeah sure, I didn't said that the cuts were too tight :) But let's imagine that for whatever reason I want to cut at 70 GeV or use only tight muons, it'd be good to have this possibility :smiley_cat: 

For customized the cut, it's ideally the solution, but for the moment, the cut value would be stored in the analyzer itself, and you need it in the category code. More decoupling would be preferable, but it's indeed a viable solution for the mean time.

plugins/DileptonCategories.cc
```diff
@@ -0,0 +1,117 @@
...
+// ***** ***** *****
+// Dilepton Mu-Mu category
+// ***** ***** *****
+bool MuMuCategory::event_in_category_pre_analyzers(const ProducersManager& producers) const {
+    const MuonsProducer& muons = producers.get<MuonsProducer>("muons");
+    return muons.p4.size() >= 2 ;
+};
+
+bool MuMuCategory::event_in_category_post_analyzers(const ProducersManager& producers, const AnalyzersManager& analyzers) const {
+    const DileptonAnalyzer& dilepton_analyzer = analyzers.get<DileptonAnalyzer>("dilepton_analyzer");
```
blinkseb: The name of the ``dilepton_analyzer`` defined in the python configuration is ``dilepton``, not ``dilepton_analyzer``.

blinkseb:
@OlivierBondu I've done a PR to your repo to fix the issue with analyzer naming.

Except that, it runs fine and seems to produce correct results :) I propose to merge as it (once the fix is included). We'll need then to start working on refinement:

* Add a way to specify the execution order of the analyzer. I though that ``cms.PSet`` kept the order used in the python configuration, but it's not the case and the analyzers are ran in alphabetic order. In this case it's working because ``dilepton`` is before ``test``, but we can't obviously rely on this :)
* Add a way to pass configuration to categories

OlivierBondu:
ok for merging then, 'hh' and 'tt' are also after 'dilepton' in alphabetical order, so it should be fine :)
I'll open issues regarding your two points so that we don't forget them

Issue #38, opened 2015-09-10
Update to CMSSW 7.4.12


blinkseb:
It seems that MVA electron IDs are merged into 7.4.12, so we should update to this release at some point. I tried and there's (at least) two issues:

- First:
```
----- Begin Fatal Exception 10-Sep-2015 15:51:29 CEST-----------------------
An exception of category 'InvalidReference' occurred while
   [0] Processing run: 1 lumi: 25 event: 6009
   [1] Running path 'p'
   [2] Calling event method for module ElectronRegressionValueMapProducer/'electronRegressionValueMapProducer'
Exception Message:
Asked for data from a PtrVector which refers to a non-existent product with ProductID 4:749
----- End Fatal Exception -------------------------------------------------
```

- Second: they (again) break the user interface without any announcement. ``pat::MET::uncorrectedPt``, ``pat::MET::uncorrectedPhi`` and ``pat::MET::uncorrectedSumEt`` are no longer existent (it's actually worst than that, these functions are still declared but the implementation is no longer present...). Replacement functions are ``pat::MET::uncorPt``, ``pat::MET::uncorPhi`` and ``pat::MET::uncorSumEt``

I guess will have to wait to a patch release.

blinkseb:
As documented here: https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookMiniAOD#Run2_Spring15_campaign_2015_vers, we can't upgrade to 7.4.12 until the next round of MiniAOD.

blinkseb:
I found a bug in CMSSW when trying to access raw MET with the new code. See
https://github.com/cms-sw/cmssw/pull/11394

swertz:
Should we bump the PRs to be merged? Especially the 74X one...

blinkseb:
I think that the 7.6.x PR must be merged before backporting. But bumping a bit on the PR would not hurt

blinkseb:
PR is finally merged for 76X. I hope 74X will follow soon

blinkseb:
Superseded by #71

PR #39, opened 2015-09-11
Migrate to CMSSW 7.4.12

blinkseb:
Mainly to not forget that I have made a branch for that.

There is an issue with electron ID crashing when running (see #38): we need to wait for an update before upgrading.


blinkseb:
Nice, build time is now 6min instead of 50min :smile:

PR #43, opened 2015-09-11
Bring some customization to dilepton analyzer

blinkseb:
Add a way to choose between loose or tight leptons, and choose which id tag to use for electrons.

### DIFF
--- a/interface/DileptonAnalyzer.h
+++ b/interface/DileptonAnalyzer.h
@@ -5,10 +5,21 @@
 #include <cp3_llbb/Framework/interface/Category.h>
 
 class DileptonAnalyzer: public Framework::Analyzer {
+    enum class WorkingPoint {
+        LOOSE,
+        TIGHT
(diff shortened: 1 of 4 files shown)


OlivierBondu:
mmm, especially for the case of electrons, since we have many eID in store, wouldn't it be easier to pass the full string of the ID ?

It would be easier, say, to apply the HEEP ID out of the box, or medium WP, or mvaID

(note to self: maybe some day we'll want to move to asymmetric IDs ?)

blinkseb:
That's already supported, see the parameters ``electrons_loose_wp_name`` and tight

OlivierBondu:
ah, my mistake, I thought https://github.com/cp3-llbb/Framework/pull/43/files#diff-e42ae69ee30d7a996836cadc17a256c6R21 was just moving around the hard-coding, instead of adding a way to specify this in the config file...

merging then !

PR #59, opened 2015-09-18
Add support for automatic PR build via Jenkins

blinkseb:
As the title says, this PR is the final touch to the support of automatic PR build via Jenkins.

[Jenkins](https://jenkins-ci.org/) is a [CI](https://en.wikipedia.org/wiki/Continuous_integration) tool to automatize all sort of things. In our case, it is used to automatically builds all the PR to check that nothing is broken.

For more details, see the updated README. If everything is working correctly, as soon as this PR is opened, the build should be triggered. Let's see!

EDIT 1: It's building!


OlivierBondu:
the build is successful !

can we remove the do not merge label ?

blinkseb:
Yes :D

PR #60, opened 2015-09-18
Support for producers scheduling

blinkseb:
As title says. This will be useful to convert some analyzers into producers.

Note: this changes the python function Framework.schedule in order to add an additional named parameter. If we follow our proposed guideline, this means we should bump major version. Do we all agree on that?

### DIFF
--- a/interface/Framework.h
+++ b/interface/Framework.h
@@ -50,11 +50,11 @@ class ExTreeMaker: public edm::EDProducer, ProducerGetter, AnalyzerGetter {
         std::string m_output_filename;
         std::unique_ptr<TFile> m_output;
         std::unique_ptr<ROOT::TreeWrapper> m_wrapper;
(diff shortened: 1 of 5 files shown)


OlivierBondu:
yes, I would agree, but we would need to decide how and when we tag a new version, not every commit I guess ?

blinkseb:
Indeed, we need a release strategy :) Tagging every commit is IMO a non-sense. Maybe we can imagine tagging every week, or before a new campaign?

OlivierBondu:
yep, every beginning on the week to tentatively rerun end-to-end things for Friday ?
or every new JSON file ?

delaere:
Note that you don't need to tag for that (weekly reruns), as every git commit is a full snapshot. You can even search the state by date. So I would limit tags to notable points: presentations, drafts, etc.

PR #62, opened 2015-09-21
Add a way to access other analyses from analyzers

blinkseb:
As the title says. A new parameter is added to the analyze function. A check is also done to see if the analyzer or producer has already produced some data when trying to access it.

This break compatibility with analysis because of the added parameter.


swertz:
Great, I was just thinking that that would be useful! Thanks!

blinkseb:
Skip tests
Skip test

PR #63, opened 2015-09-21
Add very simple unit test to check that the framework is running

blinkseb:
This PR adds a way for Jenkins to actually run the framework and see if it's working or not.

For this, it uses CMSSW unit test integration that can be launched using ``scram b runtests``.

There's only a single test for the moment, running the framework without any analysis over a ttbar MC file, stored on afs (build-bot has not certificate :sob:). Others can be easily added, more code oriented using cppunit. We just need to find someone motivated to write those tests :smiley: 

Let's see if it works!

### DIFF
--- a/test/BuildFile.xml
+++ b/test/BuildFile.xml
@@ -1,0 +1,4 @@
+<bin   name="Testcp3_llbbFramework" file="TestDriver.cc">
+  <flags TEST_RUNNER_ARGS="/bin/bash cp3_llbb/Framework/test run_tests.sh"/>
+  <use   name="FWCore/Utilities" />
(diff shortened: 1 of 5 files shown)


blinkseb:
Status should be failed since the tests failed. Investigating.

OlivierBondu:
Should we, for safety, add a simple unit test on data ?

blinkseb:
Sure, we just need to put a small data file on afs. Was too lazy to do it :)

blinkseb:
Everything's ok from my side.

OlivierBondu:
merging then ! thanks !

PR #66, opened 2015-09-23
HLT filtering and HLT objects

blinkseb:
This PR changes the way we store HLT paths. Previously, all the path were stored if they were fired for a given event. This this PR, only HLT path of *interest* are saved into the tree if they are fired. This dramatically reduces the number of paths kept in the tree, and so the output size.

The paths of *interest* are described in an XML file, where you can specify for a given run range a list of path. In this description, a path is actually a regular expression, so one can write ``HLT_.*_v.*`` if needed.

A default XML file is included which list the recommended dilepton trigger as per the Top group. Feel free to comment if you want me to add more.

Also, HLT objects are also stored if they belongs to an HLT path of interest. Only the ``p4``, the PDG id an the list of path of the object are stored.

### DIFF
--- a/data/triggers.xml
+++ b/data/triggers.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<triggers>
+    <runs from="0" to="999999">
+        <!-- Double mu -->
+        <path>HLT_Mu17_TrkIsoVVL_(Tk)?Mu8_TrkIsoVVL_DZ_v.*</path>
+        <!-- Double ele -->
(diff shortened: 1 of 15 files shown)


OlivierBondu:
The path list looks ok to me
(there are a couple of  additional ``Lepton23`` paths in the current Hww list but this is second order trouble at the moment)

thanks !

blinkseb:
The XML file can obviously be updated by the analysis to include paths of interest, using something like:

```python
process.framework.producers.hlt.parameters.triggers = cms.untracked.FileInPath('cp3_llbb/MyCoolAnalysis/data/file.xml')
```

PR #67, opened 2015-09-28
Migrate to CMSSW 7.4.13

blinkseb:
This is #39 but against the newly branch.


blinkseb:
test this please

blinkseb:
If we need to upgrade, this can be merged. I've manually backported the MET fix in the recipe.

However, I'm not sure if we can read old MC with this release or not.

src/METProducer.cc
```diff
@@ -11,7 +11,7 @@ void METProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup)
 
     sumEt = met.sumEt();
 
-    uncorrectedPt = met.uncorrectedPt();
-    uncorrectedPhi = met.uncorrectedPhi();
-    uncorrectedSumEt = met.uncorrectedSumEt();
+    uncorrectedPt = met.uncorPt();
```
swertz, line 14: Since this changed, I guess reading old MC won't work?

In principle the re-miniAOD started last Friday, using 7414 (which means we'll also have to update to this release and not 7412), however I don't know how to check that this is the case...

blinkseb, line 14: Indeed. But if `redoJEC` is true, then we recreate a new MET, so in this case it should work as long as we don't use the one from MiniAOD

PR #74, opened 2015-10-06
Add support for JEC databases

blinkseb:
A new argument is added to the `create` function. If set, it should contains the name of the database where the JECs will be read (All these database are now stored [here](https://github.com/cms-jet/JECDatabase/tree/master/SQLiteFiles)).

It's the responsibility of the analyzers to download and correctly set this argument to the value they want.

This also needs some changes in GridIn to handle the new input file (the database must be send by crab with the others files). This is handled by introducing a new _super_ variable, `process.gridin`. This variable will be read by GridIn when submitting the jobs. All the files added to the `input_files` member will automatically be sent by crab.

### DIFF
--- a/python/Framework.py
+++ b/python/Framework.py
@@ -28,7 +28,7 @@ def change_process_name_(module, process_name_from, process_name_to):
                 change_process_name_(value, process_name_from, process_name_to)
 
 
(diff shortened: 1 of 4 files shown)


blinkseb:
test this please

PR #86, opened 2015-10-17
Make packed particles branches transient

blinkseb:
Forward-port of #85


swertz:
You mean packed?

blinkseb:
Yes sorry, I just forward port the commit of #85 which is wrong too

swertz:
Check failed because of 

```
++ git fetch blinkseb
error: RPC failed; result=18, HTTP code = 200
fatal: The remote end hung up unexpectedly
```

Since #85 was OK I think this can be merged as well?

blinkseb:
Yep you can merge!

PR #90, opened 2015-11-13
SoftDrop and TopTag subjets

AlexandreMertens:
adding "Get" Btag For Subjets and For TopTag method. 
adding Subjets and TopTag input string in config file.


blinkseb:
:+1:

OlivierBondu:
:+1:

PR #93, opened 2015-11-17
Add new weight for PU reweighting

blinkseb:
As title says. I've computed the PU profile for data with latest golden JSON file, using min bias cross section of 69 nb.

The weight is stored in the 'pu_weight' branch, and computed from `EventProducer`, only if requested. The location of the root file containing the data pu profile is configurable in the python cfg.

Validation (red data, blue MC):
pu_reweighting_validation


OlivierBondu:
test this please

PR #95, opened 2015-11-20
Update b-tagging scale factors for 25ns

blinkseb:
B-tag scale factors for 25ns are here [1]. This PR remove the olds SF from Run1 and adds the new one for CSVv2 only (JP can be done also easily).

The script converting the CSV file into JSON has also been updated to fix some bug (they chose to have a different binning for nominal SF and up/down variation...)

**Note**: for the moment, any jet outside the SF boundaries will be assigned a SF of 0.

[1] https://twiki.cern.ch/twiki/bin/view/CMS/BtagRecommendation74X


swertz:
Thanks a lot! :+1:

PR #96, opened 2015-11-24
Update muon SFs for 25ns

blinkseb:
Muon SFs are here. This PR updates the SFs to the latest versions, and also fix the script responsible for converting the SFs to JSON (they also decided to change the file format for Run2 :+1: )

https://twiki.cern.ch/twiki/bin/viewauth/CMS/MuonReferenceEffsRun2

We have SF for loose, medium and tight ID, and for loose iso and tight iso (only for medium and tight id)


OlivierBondu:
Many thanks !

PR #100, opened 2015-12-07
Modifying the code to get leptons and BTagging SF=1 for data

sdevissc:
This is one possible method to get the btagging SF=1 when running on the data. The member isRealData() of edm::Event is simply used as an argument of store_scale_factor(..), and the choice of filling with the values of the SF JSONs or to use (1,0,0) is done based in the value of isRealData().


blinkseb:
Nice thanks! Can you also do the same thing for the electrons and muons?

sdevissc:
Sure I'll do that this evening.

blinkseb:
Thanks!

PR #101, opened 2015-12-08
Start of doxygen documentation

blinkseb:
I have this branch in my fork for a while.

It's a start of code documentation. Most of the classes still need to be documented, and you can use `Producer.h` as an example if you want to contribute to the effort :smile: 

Documentation is hosted on http://cp3-llbb.github.io/


OlivierBondu:
thanks !

Issue #106, opened 2016-01-07
Update to CMSSW 7.6.3


blinkseb:
7.6.x rereco are getting there, so it's time to prepare the update.

I have a branch on my fork ready to be merged: https://github.com/blinkseb/Framework/tree/CMSSW_7_6_3p

If everyone agrees, I can create a new branch here named `CMSSW_7_6_3p`, and create a PR against this branch for review.

Jenkins will also need some loves since there's a change of `SCRAM_ARCH` for 7.6.x

swertz:
That's OK with me :+1:

swertz:
The b-tag cut values as defined in https://github.com/cp3-llbb/Framework/blob/CMSSW_7_4_12p/test/TestConfigurationData.py#L35 will have to be updated, see https://indico.cern.ch/event/474580/session/1/contribution/16/attachments/1208282/1761511/BTagPerf_160106_76XWPs.pdf
I don't know if anyone actually uses these producers, but each analysis will have to change those values anyway.

blinkseb:
Indeed, but these files are only for testing if the framework is running or not, so it's really important to have up-to-date values here.

But we should definitively keep that in mind for the analyses :+1:

PR #107, opened 2016-01-07
Update to CMSSW 7.6.3

blinkseb:
See #106 and list of commits for more details.

Most noticeable things is that NoHF met is gone from MiniAOD, so I removed the associated producer. Analyses will need updates if they use NoHF met.

Let's see if Jenkins is happy with this new configuration!


OlivierBondu:
seems to work fine enough :)

MiniAODv2 should be on its way "soon" (it was supposed to be submitted 2 days ago): so far only v1 samples are produced, but not all of them will appear, and they will be invalidated as soon as v2 will appear. I would strongly suggest to wait a bit that v2 is ready before doing physics with 763+ samples.

blinkseb:
Fully agree. This is more to be ready when time comes :)

PR #112, opened 2016-01-14
Adding the Rochester correction for the muons (3 new branches in the tree).

sdevissc:
Attempt to implement the Rochester correction for the muon momentum. This is obtained from https://twiki.cern.ch/twiki/bin/viewauth/CMS/RochcorMuon . It leads to 3 entries in the tree: RochCorr_pt/eta/phi, there is no modification of the genuine muon_p4 content. Another possibility: as for the scale factors: record the correction factor instead of the newly calculated value for pt, eta and phi. To be discussed.


blinkseb:
Wha, this is some very ugly code :)

Concerning additions to the framework, if we let aside Rocherters code, please fix the following points:
- We use 4 spaces for indentations, and not tabs. Please remove _all_ the tabulations
- Indentation is wrong on some lines, please fix.

Otherwise, I'd prefer to see the p4 directly modified, instead of new branches, and a flag to enable / disable it. This way, it's transparent for the analyzers. If it's enable, we could also thing to create new branch to save the uncorrected p4, like for the met.

interface/MuonsProducer.h
```diff
@@ -41,6 +41,9 @@ class MuonsProducer: public LeptonsProducer<pat::Muon>, public ScaleFactors {
 
         BRANCH(dxy, std::vector<float>);
         BRANCH(dz, std::vector<float>);
+        BRANCH(RochCorr_pt,std::vector<float>);
+	BRANCH(RochCorr_eta,std::vector<float>);
```
blinkseb: Remove tabs and fix indentation

interface/MuonsProducer.h
```diff
@@ -41,6 +41,9 @@ class MuonsProducer: public LeptonsProducer<pat::Muon>, public ScaleFactors {
...
+	BRANCH(RochCorr_phi,std::vector<float>);
```
blinkseb: Remove tabs and fix indentation

src/MuonsProducer.cc
```diff
@@ -1,7 +1,8 @@
 #include <DataFormats/PatCandidates/interface/PackedCandidate.h>
 
 #include <cp3_llbb/Framework/interface/MuonsProducer.h>
-
+#include <cp3_llbb/Framework/interface/rochcor2015.h>
+#include "TLorentzVector.h"
```
blinkseb, line 4: Add an empty line before and after the include

src/MuonsProducer.cc
```diff
@@ -39,7 +40,21 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
         //     https://github.com/cms-sw/cmssw/blob/CMSSW_7_4_15/DataFormats/MuonReco/src/MuonSelectors.cc#L756
         dxy.push_back(muon.muonBestTrack()->dxy(primary_vertex.position()));
         dz.push_back(muon.muonBestTrack()->dz(primary_vertex.position()));
-
```
blinkseb, line 42: Why?

src/MuonsProducer.cc
```diff
@@ -39,7 +40,21 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+	TLorentzVector TLmu;
+	float qter=1.0;
+	TLmu.SetPtEtaPhiM(muon.pt(),muon.eta(),muon.phi(),0.1);	
+	if(event.isRealData()){
+		rmcor->momcor_data(TLmu, muon.charge(), 0, qter);
+		RochCorr_pt.push_back(TLmu.Pt());
+		RochCorr_eta.push_back(TLmu.Eta());
+		RochCorr_phi.push_back(TLmu.Phi());
+	}
+	else{
```
blinkseb: `} else {` please

src/MuonsProducer.cc
```diff
@@ -39,7 +40,21 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
         //     https://github.com/cms-sw/cmssw/blob/CMSSW_7_4_15/DataFormats/MuonReco/src/MuonSelectors.cc#L756
         dxy.push_back(muon.muonBestTrack()->dxy(primary_vertex.position()));
         dz.push_back(muon.muonBestTrack()->dz(primary_vertex.position()));
-
         ScaleFactors::store_scale_factors({static_cast<float>(fabs(muon.eta())), static_cast<float>(muon.pt())},event.isRealData());
+	TLorentzVector TLmu;
+	float qter=1.0;
+	TLmu.SetPtEtaPhiM(muon.pt(),muon.eta(),muon.phi(),0.1);
```
blinkseb: Spaces after comma

src/MuonsProducer.cc
```diff
@@ -39,7 +40,21 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
         //     https://github.com/cms-sw/cmssw/blob/CMSSW_7_4_15/DataFormats/MuonReco/src/MuonSelectors.cc#L756
         dxy.push_back(muon.muonBestTrack()->dxy(primary_vertex.position()));
         dz.push_back(muon.muonBestTrack()->dz(primary_vertex.position()));
-
         ScaleFactors::store_scale_factors({static_cast<float>(fabs(muon.eta())), static_cast<float>(muon.pt())},event.isRealData());
+	TLorentzVector TLmu;
+	float qter=1.0;
```
blinkseb: Spaces between `=`

src/MuonsProducer.cc
```diff
@@ -39,7 +40,21 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+		RochCorr_pt.push_back(TLmu.Pt());
+		RochCorr_eta.push_back(TLmu.Eta());
+		RochCorr_phi.push_back(TLmu.Phi());
+	}
+	else{
+		rmcor->momcor_mc(TLmu, muon.charge(), 0, qter);
+		RochCorr_pt.push_back(TLmu.Pt());
+                RochCorr_eta.push_back(TLmu.Eta());
+                RochCorr_phi.push_back(TLmu.Phi());
+	}
```
blinkseb: Remove tabs in all block and fix indentations

blinkseb:
Also, do you know it's there's a CMSSW producer which creates directly a collection of corrected muons? This would be the ideal way to do it

sdevissc:
Not sure, I checked through the Muon POG page and I only found this. Apparently, this is the exact same code used by other early Run II analyses in SMP. But I can ask to the Muon POG relevant people to make sure about this.

blinkseb:
Nah don't bother with it, it's fine

sdevissc:
"Otherwise, I'd prefer to see the p4 directly modified, instead of new branches, and a flag to enable / disable it. This way, it's transparent for the analyzers. If it's enable, we could also thing to create new branch to save the uncorrected p4, like for the met."

Originally I thought that it would make more sense to treat this corrections as we treat the SF for leptons or btag, i.e. leave it as an additional information. That avoids the need to regenerate the trees in case one needs the uncorrected value.

"Wha, this is some very ugly code :)" Thanks ;-)

blinkseb:
> Originally I thought that it would make more sense to treat this corrections as we treat the SF for leptons or btag, i.e. leave it as an additional information. That avoids the need to regenerate the trees in case one needs the uncorrected value.

It's a weight acting on a object, not on an event, so you always have the risk that, when applied, it modifies the object you select, e.g your leading muon can be degraded to the sub-leading one. Applying a weight after your selection won't take that into account, so I think it's not the right way to do it.

If you change directly the p4, this is done for free, but I agree that you'll need to regenerate the trees in case you want the uncorrected values, unless you store them in tree with the corrected ones.

> "Wha, this is some very ugly code :)" Thanks ;-)

It's not your code which is in cause, but the Rochester's one. So unless you wrote it, this was not directed to you :wink:

sdevissc:
Right, that makes sense, I'll make the change to affect the muon_p4 collection directly.

sdevissc:
Any idea why it did failed?

blinkseb:
Nop.

test this please

src/MuonsProducer.cc
```diff
@@ -16,12 +18,25 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
     const reco::Vertex& primary_vertex = (*vertices_handle)[0];
 
     double rho = *rho_handle;
-
+    rochcor2015 *rmcor = new rochcor2015();
```
blinkseb: You have a memory leak here. Move this as a class member, something like
`rochcor2015 m_rc_corrector;`

No need for it to be a pointer.

src/MuonsProducer.cc
```diff
@@ -16,12 +18,25 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
     for (const auto& muon: *muons) {
-        if (! pass_cut(muon))
+	float qter = 1.0;
```
blinkseb: What is `qter`? It should probably be configurable

blinkseb:
It builds fine, but there's still some style issues. Please remove all the tabs and fix the indentation.

interface/MuonsProducer.h
```diff
@@ -41,6 +42,7 @@ class MuonsProducer: public LeptonsProducer<pat::Muon>, public ScaleFactors {
 
         BRANCH(dxy, std::vector<float>);
         BRANCH(dz, std::vector<float>);
+        rochcor2015 rmcor;
```
blinkseb, line 49: This should be private

src/MuonsProducer.cc
```diff
@@ -16,13 +15,22 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
     const reco::Vertex& primary_vertex = (*vertices_handle)[0];
 
     double rho = *rho_handle;
-
     for (const auto& muon: *muons) {
-        if (! pass_cut(muon))
+	float qter = 1.0;
```
blinkseb: Wrong indentation. And what is `qter`? Please answer this time

src/MuonsProducer.cc
```diff
@@ -16,13 +15,22 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+        pat::Muon muoncorr=muon;
+        TLorentzVector TLmu;
+        TLmu.SetPxPyPzE(muon.px(),muon.py(),muon.pz(),muon.energy());
+        if(event.isRealData()){
+                rmcor.momcor_data(TLmu, muon.charge(), 0, qter);
```
blinkseb: Wrong indentation

src/MuonsProducer.cc
```diff
@@ -16,13 +15,22 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+        }
+        else{
+                rmcor.momcor_mc(TLmu, muon.charge(), 0, qter);
```
blinkseb: And wrong indentation

src/MuonsProducer.cc
```diff
@@ -16,13 +15,22 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+        math::XYZTLorentzVector lv(TLmu.Px(),TLmu.Py(),TLmu.Pz(),TLmu.E());
+        muoncorr.setP4(lv);
+        if (! pass_cut(muoncorr))
             continue;
-
-        fill_candidate(muon, muon.genParticle());
-
+        fill_candidate(muoncorr, muon.genParticle());
         reco::MuonPFIsolation pfIso = muon.pfIsolationR03();
         computeIsolations_R03(pfIso.sumChargedHadronPt, pfIso.sumNeutralHadronEt, pfIso.sumPhotonEt, pfIso.sumPUPt, muon.pt(), muon.eta(), rho);
```
blinkseb, line 30: There's an usage of `muon` here which should be `muoncorr`. Please switch all `muon` instances

blinkseb:
Please ping when you add / update some commits, we don't get notifications.

Otherwise, indentation is still wrong. I suggest you take some time to properly setup your text editor to avoid this in the future... We use 4 spaces for indentation, and no tabs. That's a rule we agree on since the beginning, and the only one I enforce.

Please also add an option to enable / disable Rochester corrections, false by default.

sdevissc:
Ok, I wasn't aware of this 4-spaces rule :-) I'll modify that now and add the switch for the Rochester correction.

blinkseb:
Great, thanks a lot :)

sdevissc:
I'll ask to Jerome if it is possible to install 'indent' on ingrid, which may help to make sure that all codes follow the exact same style of indentation.

interface/MuonsProducer.h
sdevissc, line 49: It seems to break seg if it is set to private (which was the first thing I tried)

sdevissc:
@blinkseb done

blinkseb:
Thanks a lot, it's almost perfect! There's an inconsistency of the usage of `muon` across the code. There's some place where `muon.pt()` is used, but it's not correct if you apply Rochester, since the `pt` has changed. All should be changed to `muontemp`

However, I'd suggest changing https://github.com/cp3-llbb/Framework/pull/112/files#diff-8466c8ba7df2bcfab5ecbe183c249628R19 to

``` C++
for (auto muon: *muons) {
```

https://github.com/cp3-llbb/Framework/pull/112/files#diff-8466c8ba7df2bcfab5ecbe183c249628R20

and here https://github.com/cp3-llbb/Framework/pull/112/files#diff-8466c8ba7df2bcfab5ecbe183c249628R32 doing

``` C++
muon.setP4(lv);
```

and then using `muon` everywhere. This way you don't need a temporary variable, `muon` is directly modified.

sdevissc:
@blinkseb FYI: this now done, and the code has been shortened a bit.

blinkseb:
Thanks, looks great now! I'd like to see all commits squashed, but I'll do it. I'm closing this one and open a new one.

Thanks again

PR #117, opened 2016-01-21
Adding dca

sdevissc:
A new variable which record the 3D compatibility (expressed in terms of resolution) between the prompt leptons candidates and the PV. Typically, if fabs(dca)<5, i.e. if the distance is less than 5 times the resolution, the compatibility can be assumed.


blinkseb:
Thanks!

blinkseb:
Can you also forward this to the 7.6.x branch?

PR #118, opened 2016-01-22
Store absolute SF errors

blinkseb:
JSON file syntax has also been extended to support 3 types of errors.
Suppose we have E +/- ΔE. Errors can be defined in the JSON, via the
"error_type" parameter, as:
- "absolute" = ΔE
- "relative" = ΔE / E
- "variated" = E + ΔE or E - ΔE

All errors will be stored as "absolute" in the output tree.

This will be needed to compute the proper uncertainty on scale factors


OlivierBondu:
test this please

PR #124, opened 2016-01-25
[74X] Electrons and muons trigger efficiencies.

BrieucF:
The efficiencies computed for the HWW group are available here [1]. They are only valid for HWW group WP recommendations [2]. NB : sometimes I see relative errors bigger than 100% which will thus lead to negative efficiencies for down fluctuated value. I am not sure whether it will be a problem or not...

[1] https://twiki.cern.ch/twiki/bin/view/CMS/HWW2015TriggerResults
[2] https://twiki.cern.ch/twiki/bin/viewauth/CMS/WW2015Variables


blinkseb:
Nice, thanks a lot :)

blinkseb:
Can you port if to 7.6.x also please :bear: ?

PR #129, opened 2016-01-26
[74X] Add a way to access producer configuration

blinkseb:
Also, do not remove 'cut' from systematics producer as this would lead
to non-consistent behavior between the nominal and systematics producers


OlivierBondu:
thanks !

OlivierBondu:
It seems we forgot to forward port this one :)

Issue #98, opened 2015-11-26
Additional jet info to store for bjet energy regression


OlivierBondu:
The list of inputs is listed on [this twiki](https://twiki.cern.ch/twiki/bin/view/CMS/HiggsToBBjetEnergyRegression):
- [x] Jet η
- [x] Jet pT
- [x] Jet transverse mass
- [x] JEC
- [x] Neutral hadron energy fraction
- [x] Photon energy fraction
- [x] SecVtxdL, the 3-d flight length of the jet secondary vertex
- [x] SecVtxdeL, Error on the 3-d flight length of the jet secondary vertex
- [x] SecVtxPt, pT of the jet secondary vertex
- [x] SecVtxM, Mass of the jet secondary vertex
- [x] SecVtxNtrk, track multiplicity of the reconstructed secondary vertex
- [x] Soft Lepton pT, transverse momentum of soft lepton candidate in the jet;
- [x] Soft Lepton pTRel, relative transverse momentum of soft lepton candidate in the jet;
- [x] Soft Lepton dR, relative distance in the η-phi space of soft lepton candidate in the jet and the jet;
- [x] Ntot, total number of jet constituents
- [x] pTLeadTrk, transverse momentum of the leading track in the jet
- [x] ρ, the average pT-density of an event.

OlivierBondu:
(I'm still waiting for the code details for the exact implementation)

giacomoortona:
LLR implementation:
https://github.com/LLRCMS/KLUBAnalysis/blob/master/test/skimNtupleInclusive.cpp#L740
(variables definitions: https://github.com/LLRCMS/LLRHiggsTauTau/blob/master/NtupleProducer/plugins/HTauTauNtuplizer.cc)

blinkseb:
Closed by #130

PR #130, opened 2016-01-27
[74X] Bjet regression

OlivierBondu:
This is work in progress for having the jets passed through the bjet energy regression as documented in [this twiki](https://twiki.cern.ch/twiki/bin/view/CMS/HiggsToBBjetEnergyRegression) (adressing issue https://github.com/cp3-llbb/Framework/issues/98)

it compiles, but for the moment crashes violently :/ In any case, style and coding comments are already welcome :smile:

### DIFF
--- a/interface/JetsProducer.h
+++ b/interface/JetsProducer.h
@@ -5,6 +5,7 @@
 #include <cp3_llbb/Framework/interface/BTaggingScaleFactors.h>
 
 #include <DataFormats/PatCandidates/interface/Jet.h>
+#include "TMVA/Reader.h"
 
 
 class JetsProducer: public CandidatesProducer<pat::Jet>, public BTaggingScaleFactors {
@@ -22,12 +23,42 @@ class JetsProducer: public CandidatesProducer<pat::Jet>, public BTaggingScaleFac
                     m_btag_discriminators.emplace(btag, &CandidatesProducer<pat::Jet>::tree[btag].write<std::vector<float>>());
                 }
             }
+            if (config.exists("computeRegression")) {
+                computeRegression = config.getUntrackedParameter<bool>("computeRegression");
+                if (computeRegression)
+                {
+                    regressionFile = config.getUntrackedParameter<edm::FileInPath>("regressionFile").fullPath();
+                    std::cout << "  -> storing bjet energy regression, with xml file " << regressionFile << std::endl;
+                    bjetRegressionReader.reset(new TMVA::Reader());
+                    bjetRegressionReader->AddVariable("Jet_pt", &Jet_pt);
+                    bjetRegressionReader->AddVariable("Jet_corr", &Jet_corr);
+                    bjetRegressionReader->AddVariable("rho", &rho);
+                    bjetRegressionReader->AddVariable("Jet_eta", &Jet_eta);
+                    bjetRegressionReader->AddVariable("Jet_mt", &Jet_mt);
+                    bjetRegressionReader->AddVariable("Jet_leadTrackPt", &Jet_leadTrackPt);
+                    bjetRegressionReader->AddVariable("Jet_leptonPtRel", &Jet_leptonPtRel);
+                    bjetRegressionReader->AddVariable("Jet_leptonPt", &Jet_leptonPt);
+                    bjetRegressionReader->AddVariable("Jet_leptonDeltaR", &Jet_leptonDeltaR);
+                    bjetRegressionReader->AddVariable("Jet_neHEF", &Jet_neHEF);
+                    bjetRegressionReader->AddVariable("Jet_neEmEF", &Jet_neEmEF);
+                    bjetRegressionReader->AddVariable("Jet_chMult", &Jet_chMult);
+                    bjetRegressionReader->AddVariable("Jet_vtxPt", &Jet_vtxPt);
+                    bjetRegressionReader->AddVariable("Jet_vtxMass", &Jet_vtxMass);
+                    bjetRegressionReader->AddVariable("Jet_vtx3dL", &Jet_vtx3dL);
+                    bjetRegressionReader->AddVariable("Jet_vtxNtrk", &Jet_vtxNtrk);
+                    bjetRegressionReader->AddVariable("Jet_vtx3deL", &Jet_vtx3deL);
+                    bjetRegressionReader->BookMVA("BDTG method", regressionFile.c_str());
+                }
+            } else {
+                computeRegression = false;
+            }
         }
 
         virtual ~JetsProducer() {}
 
         virtual void doConsumes(const edm::ParameterSet& config, edm::ConsumesCollector&& collector) override {
             m_jets_token = collector.consumes<std::vector<pat::Jet>>(config.getUntrackedParameter<edm::InputTag>("jets", edm::InputTag("slimmedJets")));
+            m_rho_token = collector.consumes<double>(config.getUntrackedParameter<edm::InputTag>("rho", edm::InputTag("fixedGridRhoFastjetAll")));
         }
 
         virtual void produce(edm::Event& event, const edm::EventSetup& eventSetup) override;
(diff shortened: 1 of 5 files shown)


interface/JetsProducer.h
```diff
@@ -41,8 +65,30 @@ class JetsProducer: public CandidatesProducer<pat::Jet>, public BTaggingScaleFac
 
         // Tokens
         edm::EDGetTokenT<std::vector<pat::Jet>> m_jets_token;
+        edm::EDGetTokenT<double> m_rho_token;
 
         std::map<std::string, std::vector<float>*> m_btag_discriminators;
+        // regression stuff
+        bool applyRegression = false;
+        std::string regressionFile = "";
+        TMVA::Reader *bjetRegressionReader = new TMVA::Reader();
```
blinkseb: You have leak here since it's not delete. I'd suggest to move the `new` part in the constructor and to switch to a `std::unique_ptr<TMVA::Reader>` to not deal with the deletion.

interface/JetsProducer.h
```diff
@@ -22,12 +23,35 @@ class JetsProducer: public CandidatesProducer<pat::Jet>, public BTaggingScaleFac
                     m_btag_discriminators.emplace(btag, &CandidatesProducer<pat::Jet>::tree[btag].write<std::vector<float>>());
                 }
             }
+            if (config.exists("applyRegression")) {
+                applyRegression = config.getUntrackedParameter<bool>("applyRegression");
+                regressionFile = config.getUntrackedParameter<edm::FileInPath>("regressionFile").fullPath();
```
blinkseb: This and below should only be done if `applyRegression` is `true`

src/JetsProducer.cc
```diff
@@ -1,21 +1,97 @@
...
+#include "DataFormats/Math/interface/deltaR.h"
 
 void JetsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup) {
 
     edm::Handle<std::vector<pat::Jet>> jets;
     event.getByToken(m_jets_token, jets);
 
-    for (const auto& jet: *jets) {
+    edm::Handle<double> rho_handle;
+     event.getByToken(m_rho_token, rho_handle);
```
blinkseb: Wrong indentation ;)

src/JetsProducer.cc
```diff
@@ -1,21 +1,97 @@
...
         if (! pass_cut(jet))
             continue;
 
-        fill_candidate(jet, jet.genJet());
-
         jecFactor.push_back(jet.jecFactor(0));
         area.push_back(jet.jetArea());
         partonFlavor.push_back(jet.partonFlavour());
         hadronFlavor.push_back(jet.hadronFlavour());
+        // Variables needed for 74X b-jet energy regression as of January 26th 2016
```
blinkseb, line 22: I guess this should be computed only if `applyRegression` is true, or do we want to compute these everytime in case we want to do it offline?

src/JetsProducer.cc
```diff
@@ -1,21 +1,97 @@
...
+            }
+        }
+        leadTrackPt.push_back(leadTrackPt_);
+        float leptonPtRel_ = -1.;
+        float leptonPt_ = -1.;
+        float leptonDeltaR_ = -1.;
+        int softLeptIdx_ = -1;
+        if (softLeptInJet.size() > 0)
+        {
+            sort(softLeptInJet.begin(), softLeptInJet.end());
```
blinkseb: `std::sort`

src/JetsProducer.cc
```diff
@@ -1,21 +1,97 @@
...
 
     edm::Handle<std::vector<pat::Jet>> jets;
     event.getByToken(m_jets_token, jets);
 
-    for (const auto& jet: *jets) {
+    edm::Handle<double> rho_handle;
+     event.getByToken(m_rho_token, rho_handle);
+
+    for (auto jet: *jets) {
         if (! pass_cut(jet))
```
blinkseb, line 14: Shouldn't the regression be done before the cut?

blinkseb:
I looked at it quickly and it looks fine, thanks! I'll test it once it's not crashing anymore :wink:  I have two questions:
- Is it correct to compute the B jets regression for all jets? I'm afraid it won't work if the jet is not a real B? In this case the correct think to do I guess is to only store the weight in the tree, and to apply it at the analysis stage.
- Shouldn't this correction be propagated to the MET also? If yes, I'm afraid it'll be rather difficult to do it properly inside the framework :disappointed:

src/JetsProducer.cc
OlivierBondu, line 22: good question... I guess the full-production timescale these days is not a show-stopper wrt doing it offline... I guess it may be more stable to have always the same entries in the tree whether or not the regression is on ? no strong opinion here

src/JetsProducer.cc
OlivierBondu, line 14: good catch !

OlivierBondu:
Those are good points, I'll try to get answers from the regression developpers... my guess answers would be:
- if it outside of the training range (aka jets from ttbar samples aka mostly b jets) the pt is unchanged, so it should be 'safe'
- the MET is not an input anymore, so at least the situation is better than for Run I... but in my understanding yes you are right, it affects the jet 4-momentum so it should be propagated to the MET, then it all depends of the size of the correction itself
- another worry that came to me while testing this: is there a well-defined way to propagate jec and jer uncertainties? In principle if these are varied before the producer is run, then we're fine ?

OlivierBondu:
so after some feedback:
- the recommended way is to use the regression after btagging, so indeed storing the weight and choosing to apply it at analysis level sounds like the correct thing to do
- for the MET it seems this use case was not considered (historically coming from Hbb searches... so the question may never have arisen before)
- I was pointed to [this Run I code](https://github.com/heppynmore/heppynmore/blob/master/python/write_regression_systematics.py#L673) about propagating jec and jer, where it looks like they correct 'only' the jet 4-momentum... Are the jet contents rescaled as well when running JEC / JER upstream (eg the soft lepton pt that is an input variable) ?

blinkseb:
> so after some feedback:
> 
>    the recommended way is to use the regression after btagging, so indeed storing the weight and choosing to apply it at analysis level sounds like the correct thing to do

Ok good to know

>    for the MET it seems this use case was not considered (historically coming from Hbb searches... so the question may never have arisen before)

Seems fair so let's put it aside for the moment, and if it's really big we can improvise at this moment :smile:

>    I was pointed to this Run I code about propagating jec and jer, where it looks like they correct 'only' the jet 4-momentum... Are the jet contents rescaled as well when running JEC / JER upstream (eg the soft lepton pt that is an input variable) ?

Only the jet p4 is rescaled upstream, and the jet content is untouched, so the internal jet variable won't be affected by the JEC / JER, only the p4 is.

OlivierBondu:
PR force-updated to rebase on top of #129, in order for the regression to be activated also when running systematics

On my end it seems to behave fine code-wise (compile, runs, do something), the PR is ready for physics validation :smile:

src/JetsProducer.cc
```diff
@@ -1,21 +1,93 @@
...
+            Jet_vtxMass = jet.userFloat("vtxMass");
+            Jet_vtx3dL = jet.userFloat("vtx3DVal");
+            Jet_vtxNtrk = jet.userFloat("vtxNtracks");
+            Jet_vtx3deL = jet.userFloat("vtx3DSig");
+            regPt.push_back((bjetRegressionReader->EvaluateRegression("BDTG method"))[0]);
+        } else {
+            regPt.push_back(jet.pt());
+        }
+
+        if (! pass_cut(jet))
```
blinkseb: Unfortunately it won't work, because as soon as a jet does not pass the cut, arrays will be desynchronized (arrays before the cut will have one entry more than those after).

It's tricky to fix, but the easiest in this case would be to move this check... before the regression. It's ok in this case since we don't rescale the p4 anymore, so after / before is the same, but before fixes the desynchronization issue.

OlivierBondu:
fixed the `pass_cut` and squashed everything

OlivierBondu:
black: without regression, red: after
[regression.pdf](https://github.com/cp3-llbb/Framework/files/108977/regression.pdf)

tested on `hh_llmetjj_allTight_btagL_csv.jj_p4[0].M()` on the radion m650 sample

on my end it looks ok :)

OlivierBondu:
updated with the small fix suggested
- new validation with more stat, both normalized to the same area (was actually missing in last plot):
  [regression.pdf](https://github.com/cp3-llbb/Framework/files/109595/regression.pdf)
- to me it looks similar to [the plot](https://twiki.cern.ch/twiki/pub/CMSPublic/HigPagDec2015/ZH_mjj.png) presented [during the jamboree](https://twiki.cern.ch/twiki/bin/view/CMSPublic/HigPagDec2015#Performance_of_b_jet_regress_AN1)

@BrieucF @sdevissc @AlexandreMertens

blinkseb:
Thanks a lot, it looks perfect. I merge as soon as jenkins is happy!

blinkseb:
Thanks :)

PR #131, opened 2016-01-27
[74X] Contain TTree memory usage

blinkseb:
This is a fix to contain TTree excessive memory usage. The solution was to force the tree to flush after a small number of events. This number was tweaked by hand to stay under 1 GB of RAM after 10000 events using the HHAnalyzer (ie ~140 events).

This should prevent jobs of being killed when running with systematics.

I've also let my debugging code to show the memory usage in case it's one day useful. It's off by default but a define can bring it back.

cc @OlivierBondu


OlivierBondu:
wow, many thanks !!

OlivierBondu:
test this please

OlivierBondu:
tested overnight on crab jobs, seems to work wonderfully, thanks again !

PR #137, opened 2016-01-30
[74X] Force trigger legs efficiency fluctuation to be between 0 and 1

BrieucF:
They provide errors allowing to have negative and bigger than 100% efficiencies once applying the down/up fluctuation.


blinkseb:
Thanks! I've made some tests, and you'll also need also clamp the resulting trigger efficiency up or down as they will sometimes be greater than 1 or less than 0.

BrieucF:
Indeed, this is again done at the analyzer level, once the fluctuation is obtain from full formula. BTW, we should maybe also do it when we extrapolate SF outside of the provided range (doubling the error).

blinkseb:
You mean efficiency here and not SF right? If so, then I agree :+1: Otherwise, SF can be greater than one, so it's not an issue in this case.

BrieucF:
Sure, I meant efficiencies, sorry for the typo ;-)

PR #138, opened 2016-01-30
[74X] Add min / max values to JSON files

blinkseb:
This allows efficiencies to be always between 0 and 1, no matter the errors.

Also, a huge error has been fixed: Up and Down errors were not always stored at the same position: for electrons and muons, it was 1: down, 2: up, but for b-tagged it was inverted. This is now fixed.

I also renamed classes to avoid using `ScaleFactor` to store efficiency, so you'll need to rename:
- `ScaleFactor` to `BinnedValues`
- `ScaleFactorParser` to `BinnedValuesJSONParser`
- and the function call of BinnedValuesJSONParser from `get_scale_factor` to `get_values`

### DIFF
--- a/data/Efficiencies/Electron_HLT_Ele17_12Leg1_TightID.json
+++ b/data/Efficiencies/Electron_HLT_Ele17_12Leg1_TightID.json
@@ -1162,6 +1162,8 @@
       ]
     }
   ], 
+  "maximum": 1,
+  "minimum": 0,
   "dimension": 2, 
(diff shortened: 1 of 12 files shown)


OlivierBondu:
could we check with @BrieucF before merging ? it seems to directly affect https://github.com/cp3-llbb/HHAnalysis/pull/74 I merged an hour ago :grin:

BrieucF:
All good, thanks!

PR #140, opened 2016-01-31
[74X] Contain memory usage, round 2

blinkseb:
This time it should be enough. Inside `TTree::Fill`, at the first basket flush, branch buffer are reallocated and the maximum size if the current size of the tree. Depending on when it occurs, it can be huge, and I see sometimes mem alloc of more than 2 GB ...

The solution is to bypass this mechanism by:
- Not calling `TTree::Fill` but directly `TBranch::Fill`
- Do the flushing manually. I've reused the value used in CMSSW

**Note: You'll need to update the `TreeWrapper` repository otherwise it won't builld!**


OlivierBondu:
**many** thanks !

PR #143, opened 2016-02-02
[74X] Framework does some fitness

blinkseb:
Output size of the tree starts to be an issue, so this PR removes some branches from the tree (they are still available on the analyzers side). Please check if there's a branch you'd like to keep on the output tree, or if there're you'd like to remove.

For 2799 events, output size is reduced to 12.3 MB from 17.5 MB (~-30%)


BrieucF:
Fine for me. Thanks. We also keep all the input variables for the btag training (vtxNtrack etc..). These branches are small but in the end they are there 5 times. Do you think we could gain by removing them?

blinkseb:
I've updated the PR to also remove the input variable for b-jet regression. They are only computed when running b-jet regression and not stored anymore in the output tree.

PR #147, opened 2016-02-10
[74X] Compute and store PDF and scale uncertainties

blinkseb:
For PDF uncertainties, this add three branches to the tree:
- `event_pdf_weight`: nominal pdf weight. Should be very close to one, but for some PDFs can be slightly different
- `event_pdf_weight_up`: up variation of the pdf weight
- `event_pdf_weight_down`: down variation of the pdf weight

PDF variation also includes alphaS uncertainties when available.

Only support for NNPDF30 is included. If a different PDF was used for the sample generation, a warning is printed to inform the user.

Scale variations are also included as a new branch, `event_scale_weights`, which will contains 6 values for the different renormalisation and factorisation scales variation (nominal is excluded, as well as anti-correlated variation as advised).

In addition, since all these uncertainties must affect only the shape and not the normalization, additional sum of event weight are stored in the final file:
- 3 for PDF variations (nominal, up and down)
- 6 for scale variations

All these new sums are named `event_weight_sum_<tag>`, where `<tag>` is either `pdf_nominal`, `pdf_up`, `pdf_down`, `scale_0` ... `scale_5`. This naming convention will be recognized by `runPostCrab.py` to store these additional weights to SAMAdhi (PRs for each repository incoming). At the end, it'll be possible to request to `histFactory` to normalize a given histogram to either the nominal case (default), or to `<tag>`, which is the same value used to fill the database.

Validation wise, PDF variations are about +- 5%, with rare outlier going up to 600%. Scale variations are a bit more important, with highest value ranging from ~+70% to ~-50%. But this will be absorbed into the normalization, so we'll need to see real effects on shape.

It'd be good if someone can run it on something else than TT (a LO sample would be great :smile:) and confirms that both tree branches and sum of event weight are (properly?) filled.

---

SAMADhi PR: https://github.com/cp3-llbb/SAMADhi/pull/11
GridIn PR: https://github.com/cp3-llbb/GridIn/pull/69

---

~~EDIT: I expect Jenkins tests to fail since I have not updated the reference trees.~~


blinkseb:
Would be good to set `pdf_weight` and friends to 1 by default (especially on data). I'll do it tomorrow

src/EventProducer.cc
```diff
@@ -59,13 +167,91 @@ void EventProducer::produce(edm::Event& event_, const edm::EventSetup& eventSetu
...
+        if (has_alphas_uncertainty) {
+            float weight_alphas_up = lhe_info->weights()[m_pdf_weights_matching[n].second].wgt / lhe_originalXWGTUP;
+            float weight_alphas_down = lhe_info->weights()[m_pdf_weights_matching[n + 1].second].wgt / lhe_originalXWGTUP;
+            // 0.75 factor is needed because NNPDF30 alphaS uncertainties are
+            // +- 0.002 whereas recommendations for Run2 are +- 0.0015
+            float alphas_uncertainty = 0.75 * (weight_alphas_up - weight_alphas_down) / 2.;
+#ifdef DEBUG_PDF
+            std::cout << "alphaS uncertainty: " << alphas_uncertainty << std::endl;
+#endif
+            rms = std::sqrt(rms * rms + alphas_uncertainty * alphas_uncertainty);
```
OlivierBondu, line 381: ah, so alpha_S uncertainties will be contained within the pdf uncertainties for NLO samples ?
(fine with me, just making a note so that people realize ;) )

blinkseb, line 381: Indeed :) If preferred, we can create a new branch for alphaS and one for the combination, but IMO it's useless since you won't apply one without the other.

blinkseb:
Ok bad new this does not work _at all_ for LO sample, since the structure of the block describing with ID is which PDF is totally different than for NLO samples ( :angry: ). I'll need some time to make it works also on LO sample ...

blinkseb:
I've updated the code to deal with both LO and NLO, so on my side it's ready for testing.

blinkseb:
please test

blinkseb:
It crashes on powheg samples :cry:

blinkseb:
Crash fixed, it should now works for everything.

&lt;rant on>
For those interested, the PDF block is different between Powheg and Madgraph:

```
<weightgroup combine="gaussian" type="PDF_variation">
```

vs

```
<weightgroup combine="gaussian" name="PDF_variation">
```

Yes, `type` becomes `name` :angry: Literally **everything** is made to make our life difficult ...
&lt;/rant on>

blinkseb:
please test

blinkseb:
So, @OlivierBondu found that, unfortunately, we are affected by the LHE generation bug for (at least) the HH signal :cry: (see [here](https://hypernews.cern.ch/HyperNews/CMS/get/generators/2648.html) for more details). This means the PDF weights we obtain are crazy:

##### Nominal value (should be 1, or very close to one)

pdf_weight_nominal_originalxwgtup pdf

##### PDF weight up

pdf_weight_up_originalxwgtup pdf

So, to fix that, I started implementing an alternative code to use LHAPDF for LO sample in order to recompute the weight for each event, by evaluating all the members of the PDF for each parton, divided by the nominal PDF value (method used for Run1 and described [here](https://twiki.cern.ch/twiki/bin/viewauth/CMS/TopSystematics#PDF_uncertainties) for example). This gives me the following distributions (same 1000 events as above)

##### Nominal value

pdf_weight_nominal_lhapdf pdf

##### PDF weight up

pdf_weight_up_lhapdf pdf

Much better, exactly what we expect (mean of 1 and about 10% uncertainties). However, it's slow (~5 times slower), and with this method we are not able to compute scale variations ...

Fortunately, there's a way to use the weights inside the LHE file. Instead of dividing by `originalXWGTUP` (as you are supposed to do if you follow the recipe, see [here](https://twiki.cern.ch/twiki/bin/viewauth/CMS/LHEReaderCMSSW#Retrieving_the_weights)), you can divide by `SCALUP`. I don't exactly know what the difference between these two values are, but, if you do that, you get:

##### Nominal value

pdf_weight_nominal_scalup pdf

##### PDF weight up

pdf_weight_up_scalup pdf

I have not cloned the distributions, it gives exactly the same results as using directly LHAPDF, but:
- It's a **lot** faster
- It's allow to also recover the scale variations

It's possible to use any of the 3 above methods in the code via some `#define` (so `originalXWGTUP`, `SCALUP` or LHAPDF). This PR use by default the `SCALUP` method, since it's a lot faster and gives exactly the same results as LHAPDF. Also note that the `#define` affects only the LO PDF / scale variation weights. Weights for NLO are always read from the LHE file since they are correct.

Hopefully it's the last fix needed!

AlexandreMertens:
Great! Is there another check to do before merging?

AlexandreMertens:
Hey, So I checked and it looks fine :).
Just one question, with no impact, but why is there a rms of 0.0006072 for the event_pdf_weight?
Well, it's small and centred at 1.

blinkseb:
Thanks @AlexandreMertens for testing :)

It's expected that `pdf_weight` is not constant, since it's the mean of all the variations of the PDF for a given event, but, as you saw, it's should be close to one with a small RMS, so that's good.

I guess if no one raise objections, we can merge this soon!

OlivierBondu:
Tested again and fine with me :)

Many many thanks for this one !

Issue #152, opened 2016-02-16
New electron ID SF for 7.4.X / 7.6.X


blinkseb:
See:
- https://indico.cern.ch/event/491499/contribution/1/attachments/1228459/1799653/CutBasedId_EGM_15Feb.pdf
- https://arun-afs.web.cern.ch/arun-afs/Final_Fits_Data_74X/
- https://arun-afs.web.cern.ch/arun-afs/Fits_Data_ID_76X/

PR #154, opened 2016-02-16
[7.4.X] Fix alphaS uncertainty computation. Closes #151

### LINKED ISSUE #151: AlphaS uncertainties are invalid
Code here is invalid : https://github.com/cp3-llbb/Framework/blob/CMSSW_7_4_12p/src/EventProducer.cc#L386-L388

It should be 1.5 instead of 0.75, see https://hypernews.cern.ch/HyperNews/CMS/get/physics-validation/2601.html

blinkseb:


OlivierBondu:
Thanks !
Shouldn't the trees be updated as well ? `pdf_weight_up` and `pdf_weight_down` will change

blinkseb:
Indeed :) I'll update them after #152 is merged

blinkseb:
Instead I rebased against #152 so we'll be able to merge both without the need to regenerate the trees.

PR #157, opened 2016-02-16
[74X] Do not recluster the jets when applying new JEC

blinkseb:
This was on my todo list for a long time.

Previously, to apply new JEC, the jets were reclustered from PF particles and the new JEC was apply as part of the usual jet clustering chain. This is obviously very slow, and it does not make real sense. Indeed, even with new JEC, the jet clustering will be the same, since JEC are applied on jets, not on PF particles.

This PR change this by simply rescaling the existing jet P4 by the right amount, and it's a lot faster.

I also fixed for good this time the fact that MVA electron ID was still ran in the background even if not used.

With all these change and running the same job, runtime drops from 15s to 7s :)


OlivierBondu:
Thanks !
Can I ask why the drop of `btagDiscriminators` in `Framework.py` ?
I assume all this reshuffling around the MET was needed by the previous way of handling the JEC ?

For once I am not sure why the ref trees need to be updated: the content should be the same ?

blinkseb:
> Thanks !
> Can I ask why the drop of btagDiscriminators in Framework.py ?

It was because, since the jets were reclustered, no b-tagging information were included, so we needed to reran also b-tagging, and all the other stuff like PU jet id, etc. It's no longer needed since we keep the MiniAOD jet, and only rescale the P4, so it keeps all the already computed quantities.

> I assume all this reshuffling around the MET was needed by the previous way of handling the JEC ?

Yep I figured out a much easier way to propagate the change in JEC to the MET than done previously. It's also much faster :)

> For once I am not sure why the ref trees need to be updated: the content should be the same ?

It should indeed, but it won't be _exactly_ the same. Jets on MiniAOD are reclustered using MiniAOD PF particles, so in 16-bits float precision, while jets during RECO are reclustered using full-precision MiniAOD. This leads to _very_ small differences in the resulting P4 of the jets.

OlivierBondu:
all clear, thanks !

PR #159, opened 2016-02-17
[74X] Switch to workaround mode for PDF weights only if needed

blinkseb:
#147 introduced PDF weights, with a workaround using `SCALUP` instead of `originalXWGTUP` for LO to compute LHE weights because of CMS generation bug affecting LO samples.

Unfortunately, this bug does not affect **all** the LO samples. For example, LO DY samples are fine...

I added a rather brute force method to detect if the workaround is needed or not: for the first event, the first PDF weight is computed. If it's greater than 100 (totally arbitrary value), then the workaround mode is used, otherwise not.

Gives expected results on both HH signal and DY LO samples.
## 

PR based over #158


BrieucF:
Thanks!

PR #163, opened 2016-02-18
[74X] Ignore pathological events when computing PDF uncertainties

blinkseb:
See https://hypernews.cern.ch/HyperNews/CMS/get/generators/2900.html for details.

The workaround used here is to discard events when one of the weight is either NaN, < 0.4 or > 2.5. This discards some events in all the samples I tested, but a very small fraction (<1%), so it should not impact the overall uncertainty estimation.

### DIFF
--- a/src/EventProducer.cc
+++ b/src/EventProducer.cc
@@ -314,7 +314,6 @@ void EventProducer::produce(edm::Event& event_, const edm::EventSetup& eventSetu
         }
 
         // PDF variations
-
         for (auto& weight: lhe_info->weights()) {
             lhe_weights.push_back(std::make_pair(weight.id, weight.wgt));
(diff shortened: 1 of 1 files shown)


OlivierBondu:
Why the 0.4 and 2.5 ?

OlivierBondu:
(love the branch name btw)

blinkseb:
TBH, because I needed to put some numbers :) I put first 0.1 and 10, but there's some events with weights like 0.2 or 5 which is obviously crazy but escaping the cut, but I have not figured out a good 'crazy scale' estimate, so I you have better values or idea! I'd even restrict it a bitter tigher, like 0.5 / 2 or even 0.67 / 1.5 ...

PR #164, opened 2016-02-22
[74X] Read JEC uncertainties from the GT by default

blinkseb:
Previously, JEC uncertainties were read from a text files, stored inside the `data` folder. The issue here is that we'll probably forget about it and fail to update this text file when new uncertainties are available.

This PR change the way uncertainties are retrieved. Now, the default is to use the Global Tag by default, but the user still have the choice to read a text files.

I've also made easier to tell the framework you want to load the JEC from a database instead of a Global Tag. There's a new method in the `Framework` class named `useJECDatabase`:

``` python
framework.useJECDatabase('Fall15_25nsV2_MC.db')
```

This affect both the JEC and the JEC uncertainties

### DIFF
--- a/python/Framework.py
+++ b/python/Framework.py
@@ -234,6 +234,25 @@ class Framework(object):
 
         return self.producers.index(name)
 
+    def useJECDatabase(self, database):
+        """
+        JEC factors will be retrieved from the database instead of the Global Tag
+        """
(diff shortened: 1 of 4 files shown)


OlivierBondu:
Thanks !

PR #167, opened 2016-03-03
[7.4.X] Make LO workaround detector code more robust

blinkseb:
Previously, only the first PDF weight was tested to switch or no to LO workaround mode. In some cases, the first weight is really small, and only the second or third is big. We now check _all_ the weights of the event before deciding if we enable or not the workaround.


blinkseb:
I need to update the reference trees because I fixed a bug upstream in the smeared jet producer. Jets was not correctly re-sorted by pT after the smearing. If affects a very few jets and more usually at low pT, but still, a jet swap is enough for the code to tell the trees are no longer identical, so I need to update the tree.

OlivierBondu:
Thanks !

PR #168, opened 2016-03-07
[74X] Protect against corrupted scale weights

blinkseb:
This PR sounds familiar... but this time for scale weights, not PDF weights. For some events, scale weights are 0, very small (e-28) or even infinite... This adds a protection against arbitrary large values of weights (greater than 10 or smaller than 0.1).

The first commits fixes the parsing of PDF set for some Powheg samples. The code now use regex instead of trying to parse the string itself.

### DIFF
--- a/src/EventProducer.cc
+++ b/src/EventProducer.cc
@@ -12,16 +12,28 @@ void EventProducer::beginRun(const edm::Run& iRun, const edm::EventSetup& eventS
 
     int32_t pdf_set_ = lheRunProduct->heprup().PDFSUP.first;
     if (pdf_set_ < 0) {
+#ifdef DEBUG_PDF
+        std::cout << "PDF set not stored inside LHERunInfoProduct. Parsing LHE headers." << std::endl;
+#endif
         // Powheg sample. Find the PDF set in the headers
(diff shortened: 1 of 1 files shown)


blinkseb:
I'll launch a crab job to ensure the bug is fixed

swertz:
Shit, just finished a large number of jobs in 76X... Does it also affect this release?

PS: Do you also have a problem and receive the github notification emails with O(1h) delay? It's really annoying...

blinkseb:
It does not seems to affect a lot of samples (only ZZTo2L2Nu for the moment, but it's probable we missed some), but it affects both 7.4.x and 7.6.x... The symptoms is infinite sum of weights for some scales. As long as you don't have any, you're fine :)

swertz:
Shit I saw that the check passed and didn't see the "do not merge" label... I thought this label prevented the merge??

blinkseb:
Ahah no it just a label it does not prevent the PR to be merged. I'm pretty sure it fixes the issue we saw so it should not matter, and otherwise I'll open a new PR this a new fix ;)

swertz:
Ok, sorry :p

blinkseb:
I confirm the bug is fixed :)

PR #171, opened 2016-03-30
Trigger efficiencies in 76

BrieucF:
Obtain from https://twiki.cern.ch/twiki/bin/view/CMS/HWW2015TriggerResults . I also added the efficiencies for single lepton in case one wants to use it. Note : now, HWW provides a leg which already takes into account the "or" with IsoMu and IsoTkMu, trigger efficiency formula will be updated accordingly.


OlivierBondu:
The test failed, probably because you did not update the reference tree for ~~data~~ MC ?

blinkseb:
Indeed. You first need to rebase against `CMSSW_7_6_3p`, and then update the reference trees to update the scale factors.

PR #173, opened 2016-03-30
[76X] Latest round of PR port

blinkseb:
This is a port of #167 and #168 to 7.6.X.  An update of the trees should not be needed, let's see if I'm right :smile:


swertz:
Yep, thanks!

PR #174, opened 2016-03-31
[76X] Update trigger efficiencies from HWW

BrieucF:
Obtain from https://twiki.cern.ch/twiki/bin/view/CMS/HWW2015TriggerResults . I also added the efficiencies for single lepton in case one wants to use it. Note : now, HWW provides a leg which already takes into account the "or" with IsoMu and IsoTkMu, trigger efficiency formula will be updated accordingly.


blinkseb:
Thanks!

PR #177, opened 2016-04-25
Ensure correct indexing of electrons. Fixes #176

### LINKED ISSUE #176: Question on edmRef in ElectronsProducer
If I understand the documentation at https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideEDMRef correctly, the index used to construct an edmRef should correspond to the index of the object in the container... isn't there a (potential) bug at https://github.com/cp3-llbb/Framework/blob/CMSSW_7_6_3p/src/ElectronsProducer.cc#L31 then that the index does not get incremented in case the electron fails the cut (because of the continue statement)?

blinkseb:
Thanks @pieterdavid for this one!


blinkseb:
Reference trees need to be updated because of (another) bug fix in the smeared jet producer.

PR #179, opened 2016-05-02
Correctly propagate JEC to MET

blinkseb:
The current code does not do what it's supposed to do unfortunately. The code is changed in order to correctly propagate the JEC to MET when redoing the JEC (ie, keeping only jet with pt > 15 GeV, skipping muons etc.)

I've checked that re-doing JEC on top of miniAOD using the same global tag leads to exactly the same pt spectrum for both MET & jets.

I'm pretty sure reference trees will need to be regenerated, but let's test jenkins with the new branch


blinkseb:
please test

blinkseb:
Reference trees updated.

blinkseb:
6 min 3 sec to build and test :smile:

OlivierBondu:
thanks !

PR #181, opened 2016-05-18
[7.6.5] Protect against MC events without weight

blinkseb:
This fix a crash on some MC samples. Some events don't have any weight stored...


OlivierBondu:
Thanks ! by 'backport needed', you mean 'forward port' right ? we don't expect to revive 74X ?

blinkseb:
I was more think about the old 7.6.X branch, but maybe it's not needed?

OlivierBondu:
it would be needed for the 80X branch but the old 76X one is not the baseline any longer ?

PR #184, opened 2016-05-24
Add support for fat jets when redoing JEC

blinkseb:
This is mainly for the ZA guys, but can be useful for everyone if we go for boosted.


OlivierBondu:
Looks like an update of the ref trees is needed !

blinkseb:
Ah crap, I though fat jets were not in any trees. So indeed, update is needed!

blinkseb:
Trees updated!

OlivierBondu:
Thanks !

PR #192, opened 2016-09-06
Read JER from GT

blinkseb:
JER factors are now stored in the GT: no need anymore to carry text files. This PR ensure we read JER factors from the GT.

I bumped both the CMSSW version and the GT to the latest available.


OlivierBondu:
Thanks ! does it mean we would just have to update the GT to run on new SF ?

blinkseb:
Yes, like the JES :)

Issue #183, opened 2016-05-22
Updates needed for 8.0.X


blinkseb:
Just a small list of what we need to update for 8.0.X. Feel free to add more stuff!
- [x] Electrons
  - [x] RECO SFs 
  - [x] effective areas
  - [x] ID
  - [x] ID SFs
  - [x] trigger efficiencies and list in `data/triggers.xml` 
- [x] Muons
  - [x] tracking SF
  - [x] SFs
  - [x] trigger efficiencies and list in `data/triggers.xml`
  - [x] muon momentum scale 
- [ ] Jets
  - [x] update jet IDs (see [the original announcement](https://hypernews.cern.ch/HyperNews/CMS/get/jet-algorithms/399.html) and [this twiki](https://github.com/cp3-llbb/Framework/blob/CMSSW_8_0_6p/interface/Tools.h#L12))
  - [ ] update PU jet ID (see [the twiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/PileupJetID)) -- Postponed until 2016 rereco
  - [x] B-tagging SFs (we also need to split between light & c/b jets)
  - [x] B-jet regression (see #178)
- [x] PU profile

OlivierBondu:
Closing this issue as this is about Moriond'17 code updates. Issue https://github.com/cp3-llbb/Framework/issues/257 has now been opened to keep track of end-of-the-year state of the art updates!

PR #193, opened 2016-09-16
update to Summer16 eID

OlivierBondu:
as per
https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedElectronIdentificationRun2

I'm guessing I may have to update reference trees ? let's try and see...

for updating https://github.com/cp3-llbb/Framework/issues/183


python/ElectronsProducer.py
```diff
@@ -9,10 +9,10 @@
             ea_R03 = cms.untracked.FileInPath('RecoEgamma/ElectronIdentification/data/PHYS14/effAreaElectrons_cone03_pfNeuHadronsAndPhotons.txt'),
```
blinkseb: Can you also update this file to `RecoEgamma/ElectronIdentification/data/Summer16/effAreaElectrons_cone03_pfNeuHadronsAndPhotons_80X.txt`?

blinkseb (changes requested):
Thanks a lot for this! Jenkins test is actually ok, I'm not sure why the status was not updated here...

OlivierBondu:
updated !

OlivierBondu:
please test

blinkseb:
Hm, something is wrong, tests should have failed, unless you updated the ref trees after your last commit? (history suggest it's not the case). One possibility is that the EA factors are actually the same between 7.6 and 8.0. If not, then there's something weird going on. Can you try locally and see if the computed isolation and EA factors for the electrons change from 7.6 to 8.0?

OlivierBondu:
yes sorry, the commit message is not clear, but I did update the ref trees in the last commit, see https://github.com/cp3-llbb/Framework/pull/193/commits/d955db00d089b07cb62666daed8d9c7dce1f8fda

blinkseb:
Ah cool! Thanks :)

blinkseb (approved):

Issue #196, opened 2016-09-26
Introduce run-dependent SF


OlivierBondu:
needed if we want to use HWW SFs for 80X

blinkseb:
IMO there's not much we can do, since the SF is applied on MC, without notion of runs. The only solution is to store all the possible scale factors (like we do for the various WP), and then randomly apply a given scale factor accordingly to the integrated luminosity of the run range.

We can maybe automatize this on the framework side, but is this what we want?

OlivierBondu:
you mean this may be more appropriately handled on the plotIt side ?

blinkseb:
I was more specifically thinking during the histFactory phase, where we have all the informations. But maybe directly into the framework would be doable, meaning only one SF to deal with instead of a bunch...

OlivierBondu:
mmm, in fact probably at a later phase could be better: if we are asked to do some plots per run-period, we would have to redo a prod for each ?

blinkseb:
Ah yes indeed...

OlivierBondu:
so we need two things:
- [x] ability to store them all at the framework step
- [ ] somehow deal with that at the histfactory / plotIt step

blinkseb:
First one we already have

BrieucF:
I am not sure I got the point here: if we decide to deal with run dependent SF in the framework (which seems to be the easiest IMO) everything would be transparent at the histfactory stage. Even if we are required to run on a specific period, the SF would automatically be the correct one, without to redo a prod, just applying a selection on run/evt number. What did I miss?

blinkseb:
There's no run number of MC, so you have to pick randomly a scale factor, ensuring you correctly probe the SF space accordingly to the integrated luminosity of each run period. This is done during the production, where you assume you know this luminosity. If you want only one period, this assumption is no longer valid.

pieterdavid:
Idea for 2: can we use the lumi block & event number (or a function of these - assuming that they are related to the seed(s) used to generate the event) to seed the random number generator for each event? That would be an elegant way to keep the "random" numbers reproducible.
If the relative lumi numbers are reliable enough (I would expect lumi calibrations to mostly change a common scale factor - at least as long as the lumi measurement does not strongly depend on pileup etc.), at the framework stage is also an option (as long as you save the run block or "virtual run number", you are still able to make per-block plots).
[...]

PR #197, opened 2016-09-26
Muons SF update

OlivierBondu:
This PR features several changes and updates, only about muons:
- the old trigger SF files hanging in the `ScaleFactor` directory have been deleted (aka: ZA-analyzers please take note and sync with the rest of us :smile: )
- the previous Muon SF files and source files have been deleted
- the new source files from both Muon POG and HWW have been added to `sources/muons`
- a new script `mergeEfficiencyFiles.py` is here to help merging the files from HWW which provides separately files for data and MC
- the old script `convertEGammaTxtFiles.py` to convert such files into FW format has been renamed (→ `convertEffTxtFileToSF.py`
- all new SF json files have been added to the framework `MuonsProducer.py`

Note: the HWW SF are now run-dependent as discussed in https://github.com/cp3-llbb/Framework/issues/196, but note that the run ranges provided are not in sync with Run2016X data-taking eras (:unamused:)

please review, I'll cross fingers that there is no typo hiding somewhere :wink:  I'll update the ref trees for merging only if I receive :+1:'s....


blinkseb:
Thanks a lot! I think we are missing tracking SF for muons:

> Please also remind to include muon tracking specific SFs covering HIP inefficiencies on mere track reconstruction, as from here (https://twiki.cern.ch/twiki/bin/view/CMS/MuonReferenceEffsRun2#Tracking_efficiency_provided_by)

OlivierBondu:
oops, I'll add them right away

in the meantime, some validation plots are available on ingrid at `/home/fynu/obondu/scratch/Higgs/CMSSW_8_0_19/src/cp3_llbb/CommonTools/scripts`

as a SF non-expert, they look more or less ok... but perhaps we should consider adding the x-coverage on the TProfile:s

blinkseb:
Thanks for the plots, nothing suspicious after a quick look!

OlivierBondu:
PR updated with tracking SF, along with README and extraction script for the provided rootfile

blinkseb:
Thanks a lot looks good! However, I just noticed here [1] that we use the absolute value of eta for the muon SFs, but tracking ones are specified using non-absolute values... Not sure if it's relevant or not.

[1] https://github.com/cp3-llbb/Framework/blob/CMSSW_8_0_6p/src/MuonsProducer.cc#L46

OlivierBondu:
checking with Hugues (muon POG) the original files contain:
- 1D integrated in pt vs eta
- 1D integrated in eta vs pt
- 2D pt vs abseta
  and 'usually' the 2D ones are the best, and in this case they are the ones we are taking

it just appears that the tracking study is uncorrelated with all that POG work and provides SF with a different binning

not sure how to detect either case automatically...  do you have suggestions ?

blinkseb:
We should probably at some point extend the format to specify which variable is expected in X or Y, but way beyond the scope of the PR. Looking at the numbers, taking eta or |eta| won't change much anyway...

OlivierBondu:
ok, adding this to the list of issues / features wanted then

I'll update the ref trees

blinkseb:
Thanks, I'll try to do it today. Don't forget to rebase too, your branch seems out of date :)

OlivierBondu:
rebased and squashed, ref trees updated

blinkseb:
Thanks a lot!

OlivierBondu:
for reference: partly addresses https://github.com/cp3-llbb/Framework/issues/183

PR #199, opened 2016-09-27
Describe what variables are used for the binning

### LINKED ISSUE #198: SF file format extension
Following the discussion at https://github.com/cp3-llbb/Framework/pull/197#issuecomment-249676952 we probably need an extension of the SF files handling, eg to specify which variables are expected in x and y (eg abseta or eta vs pt)

blinkseb:
Fixes #198 

Each JSON file now declared which variable is used for the binning. On the framework side, the user must now create an instance of `Parameters` instead of just a plain vector, and filling the needed values (like eta and pt for example). When evaluating the scale factor, the values of the variables are fetch from the `Parameters` class. If one is not found, an exception is thrown.

Note that filling `Eta` is enough for both `Eta` and `AbsEta`. We can easily add more variables in the future if needed.


OlivierBondu:
Thanks !

PR #200, opened 2016-09-27
Muons trigger eff v3

OlivierBondu:
Large PR but very few meaningful commits, hopefully the messages are explicit enough

Validation plots are available on ingrid at `/home/fynu/obondu/scratch/Higgs/CMSSW_8_0_19/src/cp3_llbb/CommonTools/scripts/Muon_HLT`


OlivierBondu:
the PR is too big for the diff to be displayed... a couple of small notes:
- we see the SiStrip inefficiencies in some validation plots, so in that sense I guess I don't see any 'unexpected' features
- in the `triggers.xml` I removed the `_DZ` part of the DoubleMuon HLT paths, this is intentional as per [Anna's slides](https://indico.cern.ch/event/547442/contributions/2281841/attachments/1326354/1991245/2016.08.24_HWW_tnp_kropiv.pdf) and her confirmation per email

blinkseb:
It looks like all the new JSON are missing the new `variables` key introduced by #199. Maybe I forgot to update one script?

blinkseb:
Concerning the `triggers.xml` file, it'd be great to have both sets for 2016 (ie the one we were using in 2015 + the new ones), with the correct run ranges.

blinkseb:
Otherwise thanks a lot! If I remember correctly, these efficiencies are not used by the framework, but directly by the analyses, so I don't think there's a need to update the reference trees here. And that's also why tests passed even if the JSON are missing the `variables` keys.

OlivierBondu:
> Concerning the triggers.xml file, it'd be great to have both sets for 2016 (ie the one we were using in 2015 + the new ones), with the correct run ranges.

I tried originally but the generating the ref trees failed for MC which looks for run number 0... I reverted back to the original '0 -> 999999' run range

OlivierBondu:
status update:
- ok, the variables were missing not because you forgot a script, but because I forgot to rebuild... so I was picking up the wrong (old) script, it should be fixed now, sorry about that
- I reintroduced 2015 paths in `triggers.xml` and also associated json and source files 
- I am still getting the following error message while running on MC, the corresponding `triggers.xml` can be found [here](https://github.com/OlivierBondu/Framework/blob/4ec22fb6e3ac2cb9828a828aa5f9a92d3c4027ab/data/triggers.xml) ; I am unsure what would be the proper fix, as anyhow we should not care with the current way the MC is setup if a given trigger path is present or not ?

```
CP3 llbb framework:
    - Running over simulation
    - global tag: 80X_mcRun2_asymptotic_2016_miniAODv2_v1
    - era: 25ns

# Conditions read from  CMS_CONDITIONS  via FrontierProd 

Smearing jets...
    Changing value of parameter producers.jets.parameters.jets (input tag) from 'slimmedJets' to 'slimmedJetsSmeared'
    Changing value of parameter producers.met.parameters.met (input tag) from 'slimmedMETs' to 'slimmedMETsSmeared'
New jets and MET collections: 'slimmedJetsSmeared' and 'slimmedMETsSmeared'

Framework configuration done.
    Producers: gen_particles, event, hlt, jets, fat_jets, met, muons, electrons, vertices
    Analyzers: 

28-Sep-2016 10:55:03 CEST  Initiating request to open file file:///afs/cern.ch/user/s/sbrochet/public/CP3/TTTo2L2Nu_13TeV-powheg_RunIISpring16MiniAODv2_reduced.root
28-Sep-2016 10:55:05 CEST  Successfully opened file file:///afs/cern.ch/user/s/sbrochet/public/CP3/TTTo2L2Nu_13TeV-powheg_RunIISpring16MiniAODv2_reduced.root

filters: 

producers: 
 -> Adding producer 'gen_particles' of type 'gen_particles'
 -> Adding producer 'event' of type 'event'
 -> Adding producer 'hlt' of type 'hlt'
HLT configuration: 
 -> Runs: [253659, 260627]
    - HLT_Mu17_TrkIsoVVL_(Tk)?Mu8_TrkIsoVVL_DZ_v.*
    - HLT_Ele17_Ele12_CaloIdL_TrackIdL_IsoVL_DZ_v.*
    - HLT_Mu17_TrkIsoVVL_Ele12_CaloIdL_TrackIdL_IsoVL_v.*
    - HLT_Mu8_TrkIsoVVL_Ele17_CaloIdL_TrackIdL_IsoVL_v.*
    - HLT_IsoMu27_v.*
 -> Runs: [271036, 999999]
    - HLT_Mu17_TrkIsoVVL_(Tk)?Mu8_TrkIsoVVL_v.*
    - HLT_Mu23_TrkIsoVVL_Ele12_CaloIdL_TrackIdL_IsoV_v*
    - HLT_Mu8_TrkIsoVVL_Ele23_CaloIdL_TrackIdL_IsoVL_v*
    - HLT_Iso(Tk)?Mu22_v.*
    - HLT_Iso(Tk)?Mu24_v.*

 -> Adding producer 'jets' of type 'jets'
 -> Adding producer 'fat_jets' of type 'fat_jets'
 -> Adding producer 'met' of type 'met'
 -> Adding producer 'muons' of type 'muons'
 -> Adding producer 'electrons' of type 'electrons'
 -> Adding producer 'vertices' of type 'vertices'

analyzers: 

Begin processing the 1st record. Run 1, Event 270939, LumiSection 1356 at 28-Sep-2016 10:55:15.846 CEST
----- Begin Fatal Exception 28-Sep-2016 10:55:16 CEST-----------------------
An exception of category 'StdException' occurred while
   [0] Processing run: 1 lumi: 1356 event: 270939
   [1] Running path 'p'
   [2] Calling event method for module ExTreeMaker/'framework'
Exception Message:
A std::exception was thrown.
Error: run 1 not found for triggers selection
----- End Fatal Exception -------------------------------------------------
28-Sep-2016 10:55:16 CEST  Closed file file:///afs/cern.ch/user/s/sbrochet/public/CP3/TTTo2L2Nu_13TeV-powheg_RunIISpring16MiniAODv2_reduced.root

---
Sum of event weight: 1

Job done in 1.232s
0 processed events, 0 selected.

=============================================
MessageLogger Summary

 type     category        sev    module        subroutine        count    total
 ---- -------------------- -- ---------------- ----------------  -----    -----
    1 Fatal Exception      -s PostProcessPath                        1        1
    2 fileAction           -s file_close                             1        1
    3 fileAction           -s file_open                              2        2

 type    category    Examples: run/evt        run/evt          run/evt
 ---- -------------------- ---------------- ---------------- ----------------
    1 Fatal Exception      1/270939                          
    2 fileAction           PostEndRun                        
    3 fileAction           pre-events       pre-events       

Severity    # Occurrences   Total Occurrences
--------    -------------   -----------------
System                  4                   4
```

blinkseb:
I was only takling about adding the path in `triggers.xml`, since they are active at the beginning of 2016. I'm not sure if we should keep the old efficiencies derived on 2015 or not. Any idea?

Concerning the MC triggers, we still perform checking if the triggers have fired or not on MC, since it can be used by analysis. The solution here is to add the triggers menu for MC using a run range of 0 -> 1.

OlivierBondu:
I think the trigger efficiency will remain the same, if we triggered on a collision in data in principle the efficiency is fixed in stone once data is recorded, and its measurement should be decorrelated from ID and ISO  ?

I'll add a dummy run range for MC with dummy generic HLT paths then

blinkseb:
That make sense indeed. My concern here is that we have efficiencies on 2015 data, but now it's 2016 data. I'm not sure if it makes a difference or not.

pieterdavid:
> I think the trigger efficiency will remain the same, if we triggered on a collision in data in principle the efficiency is fixed in stone once data is recorded, and its measurement should be decorrelated from ID and ISO ?

I am not up to date with the whole situation, but in principle that only fixes the numerator - if anything else (reco, ID, Iso etc.) that is applied before evaluating the trigger efficiency changed, this may change the trigger efficiency (through the denominator)...

OlivierBondu:
@pieterdavid : well I am no trigger efficiency expert, but yes there could be some dependence... that's why I said "should" be decorrelated. I hope this kind of effects are already taken into account into the provided uncertainties on the efficiencies... In any case we can leave the 2015 trigger efficiencies around as they are the state of the art, pending any rereco or update of the measurement (while for the ID, ISO, the baseline is to try to avoid any run dependence of any kind, so far)

OlivierBondu:
PR has been squashed and ref trees updated

data/triggers.xml
```diff
@@ -1,5 +1,19 @@
...
+    <runs from="0" to="1">
+        <!-- Double mu -->
+        <path>HLT_.*Mu.*_Mu.*_v.*</path>
+        <!-- Double ele -->
+        <path>HLT_.*Ele.*_Ele.*_v.*</path>
+        <!-- Muon + electron -->
+        <path>HLT_.*Mu.*_Ele.*_v.*</path>
+        <path>HLT_.*Ele.*_Mu.*_v.*</path>
+        <!-- Single mu -->
+        <path>HLT_.*Mu.*_v.*</path>
```
blinkseb, line 13: This will also match all dimuon and mu-e triggers. Probably fine?

OlivierBondu: I'm not sure of the exact syntax of whatever trigger menu there is in MC, nor where to find it... We can try adding an 'Iso' in front of the 'Mu' ?

blinkseb: Fine like this, thanks :)

blinkseb:
Thanks a lot!

OlivierBondu:
for reference: partly addresses https://github.com/cp3-llbb/Framework/issues/183

PR #201, opened 2016-09-28
Btag sf

OlivierBondu:
- addresses partly https://github.com/cp3-llbb/Framework/issues/183
- file names are exactly the same, so no change to the `JetsProducer.py`
- there are new files `_comb` which appears in addition of the `_mujets` one, but I am not sure if we want to use them, the [Btag twiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/BtagRecommendation80X) says 'The precision of the “comb” measurement is better than for the “mujets”, however for precision measurements on top physics done in the 2lepton channel, it is recommended to use the "mujets" one.' Opinions (@swertz, @BrieucF ?) are welcome
- ref trees updated

@blinkseb : what did you mean by '(we also need to split between light & c/b jets)' ?


blinkseb (commented):
Wha no need to tweak the convert script?? Neat!

Thanks a lot for that, looks great. Funny to see that light sfs are now binned in pt instead of eta!

blinkseb:
For the light / c b stuff. Currently we store only one Btag SF per jet, based one its flavor. If we want to do things correctly and treat light and heavy SFs as uncorrelated, the easiest to do is to save two SFs, one light and one heavy, filled according to the jet flavor

swertz:
Regarding the comb or QCD measurements, I don't remember any recent discussions in the BTV about it.... I can dig around a bit (unless Brieuc knows something), or we'll ask Adrien next time he's around!

BrieucF:
I am not aware of the difference either. Since we are not doing precision measurement, I would go for the comb one but indeed we can ask Adrien.

BrieucF:
Though the best would be to keep both and check which one gives best agreement in our phase space. How expensive is it?

blinkseb:
It's unfortunately not supported by the framework at the moment, you can only have one SF per algorithm & wp

OlivierBondu:
but both alternatives are in this PR, so it would 'just' be another prod with a different `JetsProducer.py`

blinkseb:
Duplicate the JetProducer is not a good idea concerning disk usage and runtime. If we really want to go this way (I mean mujets scale factors look perfectly fine), the proper way is to refactor the `BTaggingScaleFactor` class to support multiple scale factors.

OlivierBondu:
I guess the question is not only about the precision / agreement of the SF but about physical content: if the combined SF are making use of ttbar then we should be careful, as there might be some double counting in between btag- and ttbar-related systematics for the analyses ?

It would be good to have @acaudron 's opinion indeed :) I would suggest to keep as is the 'mujet' SF for now until we know better if it's safe enough to use the more precise 'combined' SF ?

acaudron:
I think the answer of your question is on the PR description. It all depends what's your plan: ttbar measurement (cross-section, mass or whatever) or something else like a search for HH. First case use mujet only to avoid correlation otherwise use comb to have a lower syst.

OlivierBondu:
Thanks @acaudron :)

I guess then by default it should be the 'comb' SFs, it should be fine for most cases here. 
Though maybe a word of warning to @pieterdavid depending on his exact target !

configuration and ref trees updated!

pieterdavid:
Thanks for the warning, @OlivierBondu , statistics won't allow for a "precision" measurement yet, unfortunately, and I am also not exactly looking at ttbar in the dilepton channel, but could be good to avoid the correlation in any case - I will keep it in mind and have a look at the `BTaggingScaleFactor` class (may be useful in case we end up with different SF for different data-taking periods at some point)

blinkseb:
Looks good thanks a lot for the effort!

PR #202, opened 2016-09-29
Electrons sf

OlivierBondu:
Addresses partly https://github.com/cp3-llbb/Framework/issues/183:
- adds HWW and EGM SFs
- adds HWW trigger SFs and trigger list in `triggers.xml`
- adds EGM 'RECO' SFs (in practice being 'gsf tracking' SFs)

todo:
- [x] update ref trees after https://github.com/cp3-llbb/Framework/pull/201 is merged


blinkseb:
Thanks a lot, looks great!

OlivierBondu:
PR rebased and ref trees updated

data/ScaleFactors/Electron_EGamma_SF2D_gsfTracking.json
```diff
@@ -0,0 +1,242 @@
+{
```
blinkseb, line 1: This file is corrupted. I'm not sure why the Framework accept such file, but the resulting values are kind of random.

blinkseb (changes requested):
The `gsfTracking` json file needs to be fixed. Otherwise thanks a lot!

data/ScaleFactors/Electron_EGamma_SF2D_gsfTracking.json
OlivierBondu: The original file is in the same format as others but seems to have only one bin in pt... it's probably the root cause of all this mess

OlivierBondu: got it, there is a typo in the script: https://github.com/cp3-llbb/Framework/blob/69ee67e3dc6bc773164e43d692012a9dc140ab87/scripts/extractElectronScaleFactors.py#L35, fix and PR update incoming

OlivierBondu:
PR updated with fixed json files: the bug was also removing the very first SF bin of the files... ref trees updated once more

data/ScaleFactors/Electron_EGamma_SF2D_gsfTracking.json
blinkseb: Thanks a lot for the fix!

blinkseb:
Ouch, nice catch in the end!

blinkseb (approved):

Issue #178, opened 2016-05-02
New b-jet regression available


OlivierBondu:
see [this presentation](https://indico.cern.ch/event/489926/contributions/2143570/attachments/1263188/1868300/cv-regression-hh.pdf), we would need to update !

OlivierBondu:
actually the presentation above is for 76X as as such is outdated

from Caterina the new weights are available for 80X training [here](https://github.com/vhbb/cmssw/blob/vhbbHeppy80X/VHbbAnalysis/Heppy/test/ttbar-spring16-500k-13d-300t.weights.xml), there exists also somewhere on github another version more for boosted stuff, we can contact her if we have interest

PR #203, opened 2016-09-29
Bjet regression update

OlivierBondu:
- Fix https://github.com/cp3-llbb/Framework/issues/178
- Addresses partly https://github.com/cp3-llbb/Framework/issues/183
- compiles and run, not validated yet
- no update to the ref trees should be needed as the regression is off by default

Note for reference: the [number of vertices](https://github.com/vhbb/cmssw/blob/7fa7e6a658e6bfc3463832ca79dff3197e789b38/VHbbAnalysis/Heppy/python/JetRegression.py#L63) input variable is taken from the Heppy framework, and is actually a [filtered list of vertices](https://github.com/cbernet/cmssw/blob/35f8d043c5d0dfda042f6d889e4e667201cec561/PhysicsTools/Heppy/python/analyzers/objects/VertexAnalyzer.py#L115), defined [here](https://github.com/cbernet/cmssw/blob/35f8d043c5d0dfda042f6d889e4e667201cec561/PhysicsTools/Heppy/python/analyzers/objects/VertexAnalyzer.py#L154-L162) (confirmed by Caterina over skype)

todo:
- [x] validate the regression (the easiest would be through [HHAnalysis](https://github.com/cp3-llbb/HHAnalysis) and some H(bb) signal sample, but the code needs some update after the couple of recent Framework PRs)


OlivierBondu:
Tested with the latest HHAnalyzer with `hh_llmetjj_HWWleptons_btagL_csv.jj_p4[0].M()` (meaning in the same fashion as [for the 74X version](https://github.com/cp3-llbb/Framework/pull/130)), on one file of the spin-0 M-450 signal, blue is without the regression and red is with the regression
- both plots normalized to the same number of entries: [80X_bjet_regression_norm1.pdf](https://github.com/cp3-llbb/Framework/files/511294/80X_bjet_regression_norm1.pdf): the mean is closer to 125 GeV
- not normalized: [80X_bjet_regression.pdf](https://github.com/cp3-llbb/Framework/files/511293/80X_bjet_regression.pdf): the regression brings in about 27% more events

looks ok to me, I'm removing the 'do not merge' label, comments are very welcome

OlivierBondu:
please test

blinkseb:
Thanks a lot!

PR #204, opened 2016-10-03
Muon scale corrections

OlivierBondu:
This PR addresses partly https://github.com/cp3-llbb/Framework/issues/183

list of changes:
- update the Rochester moment scale correction to the newest version (see [their twiki](https://twiki.cern.ch/twiki/bin/view/CMS/RochcorMuon))
- introduce the Kalman Muon Calibrator (see [their twiki](https://twiki.cern.ch/twiki/bin/view/CMS/MuonScaleResolKalman))
- the prescription for related muon scale systematics is not super clear at the moment, so nothing is implemented in this regard

Of course both shouldn't be used at the same time... 

For the review: 
- you can discard the first commit, which is just dropping old rochcor2015 files and replacing with rochcor2016 ones, the other two are the meat of the changes
- no update of the reference trees since these corrections are off by default


README.md
```diff
@@ -30,6 +30,7 @@ git clone -b CMSSW_8_0_6p -o upstream git@github.com:cp3-llbb/Framework.git cp3_
 
 # Stuff not yet in central CMSSW:
 git cms-merge-topic ikrav:egm_id_80X_v1 # 8010+ electron ID WPs
+git clone -o Analysis https://github.com/bachtis/analysis.git -b KaMuCa_V4 KaMuCa # KalmanMuonCalibrator
```
blinkseb: No tag or anything more stable? Any update upstream will change our results, which mean hardly reproducible results

interface/MuonsProducer.h
```diff
@@ -47,7 +72,8 @@ class MuonsProducer: public LeptonsProducer<pat::Muon>, public ScaleFactors {
         BRANCH(dxy, std::vector<float>);
         BRANCH(dz, std::vector<float>);
         BRANCH(dca, std::vector<float>);
-        rochcor2015 rmcor;
+        rochcor2016 *rmcor;
+        KalmanMuonCalibrator *kamucacor;
```
blinkseb: Please switch to `std::unique_ptr` and drop the destructor

src/MuonsProducer.cc
```diff
@@ -21,7 +21,28 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
         if(applyRochester){
             float qter = 1.0;
             TLorentzVector TLmu(muon.px(),muon.py(),muon.pz(),muon.energy());
-            (event.isRealData())?rmcor.momcor_data(TLmu, muon.charge(), 0, qter):rmcor.momcor_mc(TLmu, muon.charge(), 0, qter);
+            if (event.isRealData())
```
blinkseb: `{` on the same line

src/MuonsProducer.cc
```diff
@@ -21,7 +21,28 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+            {
+                rmcor->momcor_data(TLmu, muon.charge(), 0, qter);
+            } else {
+                int ntrk = 0;
+                if (!muon.innerTrack().isNull())
```
blinkseb: Same here

src/MuonsProducer.cc
```diff
@@ -21,7 +21,28 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+                {
+                    ntrk = muon.innerTrack()->hitPattern().trackerLayersWithMeasurement();
+                }
+                rmcor->momcor_mc(TLmu, muon.charge(), ntrk, qter);
+            }
+            muon.setP4(math::XYZTLorentzVector(TLmu.Px(),TLmu.Py(),TLmu.Pz(),TLmu.E()));
```
blinkseb: Space after comma

src/MuonsProducer.cc
```diff
@@ -21,7 +21,28 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+        }
+        if (applyKaMuCa)
```
blinkseb: `{` on the same line

src/MuonsProducer.cc
```diff
@@ -21,7 +21,28 @@ void MuonsProducer::produce(edm::Event& event, const edm::EventSetup& eventSetup
...
+        {
+            TLorentzVector TLmu(muon.px(),muon.py(),muon.pz(),muon.energy());
```
blinkseb: Why the `TLorentzVector` here? You can use `muon.scaleEnergy()` to rescale the P4

blinkseb (changes requested):
Thanks a lot! Good to see that Rochester code is still as ugly as before!

I have some comments, mostly cosmetics. I'm also wondering if it's possible to improve the configuration of the muon producer without the need for the _hardcoded_ code in `Framework.py`. What about adding a new argument to the `addProducer` method to allow the user to pass a custom function to be called once the producer is created, with one argument `isData`. We would then be able to move the KaMuCa configuration block into this new function, located inside `MuonProducer.py`. What do you think?

Otherwise, thanks a lot!

README.md
OlivierBondu: I just dumbly took the setup instructions, we can switch to their latest commit of today ?

OlivierBondu:
I just learned I did not apply KaMuCa for MC in the correct order, please don't merge before I remove the label :)

README.md
blinkseb: Should be enough thanks :+1:

src/MuonsProducer.cc
OlivierBondu: ha, correct, just out of laziness I copied the rochcor syntax ;)

OlivierBondu:
PR updated

OlivierBondu:
PR updated with requested changes + storing the value of the correction factor

I did update the reference trees since this is now adding one branch

interface/MuonsProducer.h
```diff
@@ -47,7 +68,10 @@ class MuonsProducer: public LeptonsProducer<pat::Muon>, public ScaleFactors {
         BRANCH(dxy, std::vector<float>);
         BRANCH(dz, std::vector<float>);
         BRANCH(dca, std::vector<float>);
-        rochcor2015 rmcor;
+        std::unique_ptr<rochcor2016> rmcor;
+        std::unique_ptr<KalmanMuonCalibrator> kamucacor;
+        BRANCH(rochester_correction, std::vector<float>);
+        BRANCH(kamuca_correction, std::vector<float>);
```
blinkseb: One branch is enough since you explicitly check if both method are used at the same time. Something like `scale_correction_factor`?

blinkseb (changes requested):
One last fix and it's good to go :+1:

interface/MuonsProducer.h
OlivierBondu: corrected in the latest commit

interface/MuonsProducer.h
OlivierBondu: done

src/MuonsProducer.cc
OlivierBondu: done

src/MuonsProducer.cc
OlivierBondu: done

src/MuonsProducer.cc
OlivierBondu: done (also other places)

src/MuonsProducer.cc
OlivierBondu: done

blinkseb (approved):

blinkseb:
Thanks a lot!

PR #207, opened 2016-10-05
Fix b-tag scale factors

blinkseb:
Format of CSV file changed a bit, with addition of new `up_correlated`, `down_correlated`, `up_uncorrelated` and `down_uncorrelated` syst types, in addition to the mandatory `up` and `down`. The convert script was not checking for unsupported type, which led to buggy and/or fishy JSON files.


OlivierBondu (commented):
Thanks a lot ! so we don't care about decorrelated / correlated errors ?

blinkseb:
According to the twiki, it's the breakdown of the syst uncertainties, needed if you start doing fancy stuff.

PR #208, opened 2016-10-05
updating PF jet ID

OlivierBondu:
partly addresses https://github.com/cp3-llbb/Framework/issues/183

PU jet ID update would be easier done in 8020 (see [twiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/PileupJetID)), so I'll separate the two


blinkseb:
Cool, let's bump to 8.0.20 then! Can you rebase and update the trees please?

OlivierBondu:
rebased & updated ref trees

blinkseb:
Thanks!

PR #210, opened 2016-10-10
Move command line parsing to a standalone class

blinkseb:
Analyses can use the same class to parse command line arguments. This fix the issue in https://github.com/cp3-llbb/HHAnalysis/pull/100#discussion_r82174372 where an exception is thrown because a command line parameter is not registered.

The fix is to use the same parser across all the code. This makes sure that all possible options are registered.


OlivierBondu:
Thanks !

PR #211, opened 2016-10-10
Split b-tagging SFs for light and heavy flavors

blinkseb:
We should treat light and heavy flavors b-tagging SFs as uncorrelated. To achieve this, this PR split the SFs branch for b-tagging into two, one for light flavors and one for heavy flavors. If a jet is light, the heavy branch is filled with a 1 +- 0 SF, and vice versa. This allow one to use the heavy SFs for every jet without knowing if it's actually a light or a heavy jet. In case this information is needed, a new branch is added, `jet_systFlavor`, which can be either 1 for heavy or 2 for light jets.

I run it and it works as expected. All SFs from the heavy branch are 1 for light jets (and the opposite), and sane values are present for light & heavy jets.

~~There's also an interface change in the `get_scale_factor` method, which now expects a `Flavor` argument, needed to decide between the light & heavy SFs.~~ No longer relevant

TODO:
- [x] Update reference trees


OlivierBondu (commented):
Many thanks ! I have only one minor comment: wouldn't it be better to have the string saying `lightjet` (resp `heavy`) instead of the plural `lightjets` ?

blinkseb:
Yes sure, it was to be consistent with the already in place strings in the configuration file (`bjets`, `cjets` and `lightjets`). I can drop the `s` :)

blinkseb:
Updated dropping the `s` from strings :wink: I also added a new method in `JetProducer` to avoid changing the interface for analyses. It's now transparent.

OlivierBondu:
Many thanks !

PR #212, opened 2016-10-14
Add HLT filters

OlivierBondu:


blinkseb:
What's the impact on the file size?

OlivierBondu:
good point: not negligible

```
Statistics for tree /t
******************************************************************************
*Tree    :t         : t                                                      *
*Entries :       68 : Total =     704 kB  File  Size =     156 kB  ~ 95.49 % *
*        :          : Tree compression factor =   3.98                       *
******************************************************************************
branch hlt_object_filters                                               21 kB    ~13.36 %
branch jet_p4                                                            9 kB    ~5.73 %
branch muon_p4                                                           4 kB    ~2.82 %
branch hlt_object_p4                                                     3 kB    ~2.13 %
branch dilepton_mumu                                                     3 kB    ~1.68 %
branch jet_puJetID                                                       2 kB    ~1.56 %
branch jet_regPt                                                         2 kB    ~1.50 %
branch jet_jecFactor                                                     2 kB    ~1.47 %
branch jet_pfCombinedInclusiveSecondaryVertexV2BJetTags                  2 kB    ~1.17 %
branch electron_p4                                                       2 kB    ~1.05 %
(...)
```

however it is needed if we want to be in sync with T&P trigger matching... and is an analysis decision, not a framework-one, so I'm not sure which way to go here

OlivierBondu:
Moved the branch to be transient: I cannot foresee a use case where we would have to access it outside of the analyzer...

This way the info still can be used, but is not stored

blinkseb:
Thanks for checking! Looks good to me, thanks!

PR #215, opened 2016-11-07
Remove cuts on PDF and scale weights

blinkseb:
Title says everything. I never liked the arbitrary cut I added on the weights to prevent them to go to big, so this PR removes them, only keeping the check against NaN. I also added a protection against negative down-variation of pdf weights, which can append if the RMS is greater than the mean value. There's also an additional protection when computing this RMS, to ensure that at least 80% of the weights are valid (this 80% is also arbitrary, but it should prevent events with a lot of NaN weights).

I'm also wondering if we should start to apply `event_pdf_weight` for the nominal plots. It's always close to one, but for some event can be significantly different. If we don't apply it, for some event we overestimate the PDF uncertainty. Not a big deal I guess, but since everything is here...

Some plots. Blue is without this PR, red with this PR.

event_scale_weights

event_pdf_weight

event_pdf_weight_up

event_pdf_weight_down

And now the effect on the jet pt spectrum (no selection on the jets, TT fully leptonic sample). 

jet_pt_event_scale_weights

jet_pt_event_pdf_weight_up

jet_pt_event_pdf_weight_down


blinkseb:
We need to solve this this week while Olivier is here

PR #216, opened 2016-11-08
Support for weight-averaged scale factors and efficiencies

blinkseb:
This PR adds support for weighted binned-values, more specifically to weighted scale-factors, needed for HWW working points for both electron & muon SFs. The integrated luminosity is used as a weight to describe how much a given set of values must be sampled. In the end, only one scale factor will be stored in the tree, equal to the weighted average of the various scale-factors for a given sample.

Note: the seed is hardcoded in order to have reproducibility, but this also means different samples will have the same suite of random numbers. If someone has an idea a way to guarantee reproducibility but with randomness across samples, please speak up :+1:


swertz:
Now that I see this, wouldn't it be possible to store an _actual_ weighted average of the different scale factors, instead of drawing one at random?

BrieucF:
If you have access to the input file name, we could define the seed according to an hash of the input file name?

blinkseb:
@swertz I'm not sure to understand, it's already what is done here? Or you mean computing the average once for all and use the result as input?

@BrieucF AFAIK this is not something we have access to

swertz:
Yes, that's what I mean...

blinkseb:
It's possible yes and probably easier. I have something working locally, I will append it to this PR, which seems to be anyway needed for the electrons, since they **decided to NOT use the same binning for the SFs across  different era :angry:**

blinkseb:
PR updated:
- new script to create weighted average of a set of JSON files
- weighted average for muons SFs and triggers efficiencies

~~Electrons still use the random method since the binning is different between the two files. Ideally we should try to create a common binning between the two files, but I can't figure (yet) an automated way to do that easily...~~

blinkseb:
PR updated again. The script creating the weighted efficiencies now find a common binning in case it's not the same between the files and compute the weighted average based on this new binning. Consequently, there's now a single file also for the electrons SFs!

blinkseb:
I just checked and using the random method or the weighted average file give exactly the same result on the muon pt spectrum (~ 20000 events) :+1:

blinkseb:
I removed the dependency on #215, this is good to merge!

PR #218, opened 2016-11-15
Store pointers to selected electrons

blinkseb:
Useful to grab some information later from the object
without having to store them in the producer

TODO: Do the same for all others producers

Note: this is already used in https://github.com/cp3-llbb/HHAnalysis/pull/110


OlivierBondu:
thanks!

PR #219, opened 2016-11-23
Add HLT-safe electron id

blinkseb:
See https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedElectronIdentificationRun2#HLT_safe_selection_for_2016_data


OlivierBondu:
ref trees needs to be updated as there is new content!

blinkseb:
Yep I'm on it :)

PR #221, opened 2016-11-28
Externalize muons correction

blinkseb:
This PR move the muons correction out of the muon producer to a dedicated CMSSW producer, as it's already done for JES and JER. The main advantage here is that now the Framework muon producer does not longer care if the muons are corrected or not, since it's done automatically before. This is controlled by the new Framework method `applyMuonCorrection(type)`, where `type` is `kamuca` or `rochester`.

This is also a move forward to anticipate systematics on these methods if we decide to apply one of them, which will be handled exactly the same as JES/JER.


OlivierBondu:
Thanks ! looks good to me, but Jenkins complains about compiling :wink: 

seems related to the `scale_correction_factor` trying to be filled while it is of no use any longer (`MuonsProducer.cc:41`)

blinkseb:
Thanks for noticing, not sure how I missed that! I've updated the PR, let's see if it builds now :+1:

OlivierBondu:
And of course, now trees need to be updated :smile: 
```
Branches found in reference tree but not in output tree:
  muon_scale_correction_factor
```

blinkseb:
Ahah, I though I could escape that :sweat: Will do tomorrow!

OlivierBondu:
Thanks !

PR #222, opened 2016-12-08
Save cMVAv2 b-tag discriminant

blinkseb:
Save cMVAv2 b-tag discriminant.

Note: scale-factors still need to be included


OlivierBondu:
It seems that (as expected) an update of the ref trees is needed.. but the actual jenkins output is weirder, it mentions some branches both in `found in output tree but not in the reference tree` AND in `found in reference tree but not in output tree`, is there an ordering condition when looking for difference between trees ?

blinkseb:
Looks like the debug output is buggy, I'll look at it

blinkseb:
Fixing the bug leads to the following error:

```
Branches found in reference tree but not in output tree:

Branches found in output tree but not in the reference tree:
  fatjet_pfCombinedMVAV2BJetTags
  fatjet_softdrop_pfCombinedMVAV2BJetTags
  fatjet_softdrop_puppi_pfCombinedMVAV2BJetTags
  jet_pfCombinedMVAV2BJetTags

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Much more logical :)

OlivierBondu:
Thanks !

PR #224, opened 2016-12-13
Store electron MVA id

blinkseb:
Compute and store electron mva id.

I except Jenkins to fail, but I first want to see if the compilation is successful. Afterward, I'll update the reference trees.


blinkseb:
jenkins please test

blinkseb:
jenkins please test again

OlivierBondu:
Thanks !

PR #225, opened 2016-12-15
Add support for electron regression

blinkseb:
This PR adds support for electron regression in the framework. Simply call `framework.applyElectronRegression()` to enable it.

See this twiki page for more information: https://twiki.cern.ch/twiki/bin/view/CMS/EGMRegression

Some validation:
mll_with_regression
M_ll distribution for loose electrons, with regression (red) and without (blue)


OlivierBondu:
thanks !

PR #227, opened 2017-01-06
Remove code that overwrites electron-R03 effective area configuration

pieterdavid:
I'm not an expert (and I didn't even run this to confirm my suspicion), but it looks like these lines in Framework.py will still set the file with effective areas from ElectronProducer.py (which was updated to Summer16) back to one of the old (Spring15 25ns/50ns) ones.


blinkseb:
Thanks a lot! Indeed, values are overwritten... But in the end, I don't think anyone use these values, since the electron ID includes the isolation.

Can you regenerate the reference trees please? Simply run the script inside `test` (it's only working on lxplus)

pieterdavid:
Done, thanks for having a look (I wasn't too worried about it, but it caught my attention while going the changes for 2016 data, so better to just fix it)

PR #229, opened 2017-01-17
ICHEP b-tagging SFs for cMVAv2

blinkseb:
Since it looks like we'll have to go to pre-app with ICHEP SFs, here we go.


OlivierBondu:
Thanks !

PR #230, opened 2017-01-24
Reject fake muons

blinkseb:
This PR adds support for the newly introduced fake muon filter. Unfortunatly, it requires to bump the release to 8.0.25... 

Implementation is straightforward: the filter is run first, and reject all the events where there's bad muons. I ran over 1000 events on DoubleMuon, run F, and 44 events were rejected.

### DIFF
--- a/CMSSW.release
+++ b/CMSSW.release
@@ -1,1 +1,1 @@
-CMSSW_8_0_24_patch1
+CMSSW_8_0_25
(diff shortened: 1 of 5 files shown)


blinkseb:
I ran a production with the filter and effect is small (< 1%), so it's good to merge!

swertz:
Thanks a lot!

PR #231, opened 2017-01-24
Bug fixes on b-tagging scale-factors evaluation

blinkseb:
B-tagging scale-factors are provided as TFormula, which we should evaluate with the jet pt to obtain the SF. These SFs are valid only for a given pt range. This PR makes sure we don't evaluate the SFs using a pt which is non valid (ie, outside the range).

I also fixed the validity range of the discriminant. It's by default between 0 and 1 inside the CSV file, but it's not correct since the discriminant can be negative. It has not real effect since the formula *does not* depend on the discriminant value, but any jet with a discriminant outside the range would cause the error to be double (standard treatment). I put arbitrarly -100 - 100 to be safe

--

Based over #230 . Do not merge before #230 is merged and this rebased.


blinkseb:
Reference trees updated

swertz:
Good to go too?

blinkseb:
Let me rebase first :)

blinkseb:
Done :+1: This should be good to merge, no need to wait for jenkins

PR #232, opened 2017-01-27
Protect against corrected PDF or scale weights

blinkseb:
Latest episode of the endless story of corrupted PDF and scale weights.

On some samples, PDF and / or scale weights, coming directly from the
LHE, are corrupted (either 0, NaN, Inf, or very small / big). This PR tries to catch such weights, by checking if the value is < 0.01 or > 100. This should be more than enough for a regular weight, and will prevent propagation of inf or NaN to the sum of weights.

### DIFF
--- a/interface/EventProducer.h
+++ b/interface/EventProducer.h
@@ -12,7 +12,10 @@
 #include <map>
 
 // Uncomment to debug PDF uncertainties
-// #define DEBUG_PDF
+//#define DEBUG_PDF
+ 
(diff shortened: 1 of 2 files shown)


swertz:
Thanks! Just something: do we care if the weight is 0 (or very small)? That will not cause any problems afterwards right?

blinkseb:
I can't think of any issue right now, but it'll probably bite us at some point, so I prefer to be safe here :)

swertz:
Fine with me ;)

PR #235, opened 2017-02-02
updating egamma regression to v2

OlivierBondu:
Thanks to @RemKamal for the notice!

If I read correctly [Framework.py#L433-L441](https://github.com/cp3-llbb/Framework/blob/CMSSW_8_0_6p/python/Framework.py#L433-L441) this should be the only update needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,8 @@ git checkout 2ad38daae37a41a9c07f482e95f2455e3eb915b0
 popd
 
 # Electron regression
-git cms-merge-topic rafaellopesdesa:Regression80XEgammaAnalysis
+# https://twiki.cern.ch/twiki/bin/view/CMS/EGMRegression
+git cms-merge-topic rafaellopesdesa:Regression80XEgammaAnalysis_v2
 
 # Fake-muon filter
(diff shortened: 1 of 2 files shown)


pieterdavid:
cms-git question: does this mean we have to re-create the release area (and possibly move to 8_0_26?), or can this be applied on top of an existing one?

OlivierBondu:
I could do the `cms-merge-topic` on top of my existing installation... no idea if that's the correct thing though

blinkseb:
If the new branch was branched from the old one, re-running the `merge-topic` should just pull the latest changes, no need for a new release

pieterdavid:
I don't see the old branch on github, but merge-topic goes fine and the merge commit looks ok as well, so I guess that's the case - thanks for explaining

RemKamal:
Hello all,

I guess Fabrice already answered, but still for completeness, doing merge on top with v2 should do the work. 
This is the main Twiki:
https://twiki.cern.ch/twiki/bin/view/CMS/EGMRegression  
and Rafael recommends to check that we are up-to-date with the instructions, which also changed at some time.

Rami

blinkseb:
Thanks :+1:

PR #239, opened 2017-02-08
Moriond17 Rochester corrections

blinkseb:
Update Rochester to latest version. I removed the old implementation in the process.

Red is with Rochester, Blue is without:
mll_rochester


OlivierBondu:
miam, all those txt files...

thanks !

PR #240, opened 2017-02-09
Add support for JEC sources when doing systematics

blinkseb:
If the flag is turned on, one jet producer per JEC source (~25*2)
is added to the framework, in addition of the usual two (up and down). You can see in `TestConfigurationMC.py` how to turn the flag on (`splitBySources`)

Not directly related but used in HH on an upcoming PR is a way to enable a branch only when running a producer or analyzer over the nominal collection and not over systematics (`ONLY_NOMINAL_BRANCH`)

It's a bit ugly, but it works...


swertz:
Awesome thanks! Just a thing: don't we want to keep the "simple" up/down JEC parametrisation even when asking the full thing? For all the control plots we don't want the breakdown...

blinkseb:
It's always done whether or not you split by sources :)

OlivierBondu:
looks good to me, but jenkins complains it lost `jet_jecdown_p4.fCoordinates.fPt` on the way... not sure why, since the code should not be turned on by the default config...

blinkseb:
That should not happen, I'll look into it

blinkseb:
Ah no it's simply because I updated the GT, so the uncertainties are different. I'll revert back to the old one

OlivierBondu:
thanks !

blinkseb:
For reference, an update of the trees was needed because I changed the way the JEC uncertainties are propagated to the jet collection. Instead of using the producer include within CMSSW, I rewrite one to be able to produce multiple collections, one per JEC source. But, it looks like the *standard* producer does **not** sort the resulting collection by pt (which is rather silly), but our new producer does.

PR #241, opened 2017-02-13
Use correct file for preliminary EGM scale

swertz:


blinkseb:
Why this change? I think the one I put is the correct one?

blinkseb:
If you look at the commit adding the `Moriond17_JEC`, it looks like it's the one targeting the full 2016 stat: https://github.com/shervin86/cmssw/commit/34327fd6063ea2b4cae079872211a8c9dbe0c4f6

swertz:
Yes sorry, I didn't re-do a `git cms-merge-topic` of all the stuff added on top of the framework...

PR #242, opened 2017-02-13
Fix various crashed related to Rochester and Smearing

blinkseb:
See commits for more details. In summary:
  - Rochester SFs are sometimes NaN if the input muons is crazy (pt > 13 TeV, stuff like that which seems to exist in standard MiniAOD collections)
  - Electron smearing crash if eta > 2.5. Twiki asks to add a filter to keep only electrons with eta < 2.5, but looking into the code, it's actually supercluster eta which is used, not eta, and I found a case where eta < 2.5 but supercluster eta = 3.81 (I know...)


swertz:
Thanks a lot for the quick fix! 

This thing with 13 TeV pt muons is scary, shouldn't we report it?

PR #245, opened 2017-02-14
Add a way for users to control tree flushing

blinkseb:
This PR adds two new options in the configuration file to control the flushing of the output tree. It can helps for reducing the memory pressure to increase the flushing operations (by decreasing the memory needed to trigger a flush). The default is still set to 15 MB but it can now be override by the users:

```
process.framework.treeFlushSize = cms.untracked.uint64(5 * 1024 * 1024)
```

I also added a missing call to `OptimizeBackets` to mimic what is done in the `TTree::Fill` method. It should also helps to better optimize the memory usage.


swertz:
Thanks! Do you have any recommendation for changes at the analysis side?

OlivierBondu:
Thanks !

blinkseb:
No recommendations for the moment, I'm trying to see what is the real effect of reducing the flush size...

PR #246, opened 2017-02-15
Upgrade to CMSSW 8.0.26p2

blinkseb:
Title says everything. Please do not merge until EGM smearing is updated (the recipe will probably change)

Note: bad muons filter has been removed since user are now expected to run over reMiniAOD


README.md
```diff
@@ -28,45 +28,25 @@ git cms-init
 git clone -o upstream git@github.com:blinkseb/TreeWrapper.git cp3_llbb/TreeWrapper
 git clone -b CMSSW_8_0_6p -o upstream git@github.com:cp3-llbb/Framework.git cp3_llbb/Framework
 
-# Stuff not yet in central CMSSW:
-# 8010+ electron ID WPs
-git cms-merge-topic ikrav:egm_id_80X_v2
```
swertz, line 33: This has been merged into CMS? The twiki still mentions the private branch...

blinkseb: Yep I checked it

swertz:
Thanks! I expect the build failed because you're waiting for the new scales to regenerate the ref trees? But the link to jenkins seems dead...

README.md
swertz: Ok :)

blinkseb:
The build should not failed since the recipe is basically the same. But:
 - 8.0.26 patch 2 is not present on AFS, that's why the build failed
 - I'm in the process of migrate the build-bot the use CVMFS instead of AFS, but it's a bit more complicated than expected...

blinkseb:
please test

blinkseb:
please test

blinkseb:
please test

blinkseb:
please test

blinkseb:
please test one more time :man_dancing:

blinkseb:
please please please test test test :rocket:

blinkseb:
:champagne: :tada: :balloon:

blinkseb:
It'd good also to implement recommendations from https://twiki.cern.ch/twiki/bin/view/CMSPublic/ReMiniAOD03Feb2017Notes

swertz:
Seems like we do have to change the MET input tag if we want to propagate the EGM fix?  => `slimmedMETsMuEGClean`

blinkseb:
Pff indeed

swertz:
Ah but we don't really care about that one: `jets are not e/gamma corrected`
=> it makes sense to use only the muon-corrected MET, which *is* the default collection?

swertz:
For the other things did you mean the ` It would be extremely advisable to store the bool particleFlowEGammaGSFixed:dupECALClusters and if ecalMultiAndGSGlobalRecHitEB:hitsNotReplaced is empty or not `?

blinkseb:
Yep that ones

blinkseb:
I'm looking also into updating MET filters

blinkseb:
please test

---

I'm trying to see if using webhooks to jenkins works with the new container infrastructure (instead of plain old cron task)

---

And looks like it's working :+1:

blinkseb:
The EGMSmearing twiki has been updated with a new recipe I implemented in ab4c4bb :tada: 

---

There's something wrong with the recipes, it pulls all packages of CMSSW instead of just the modified one... investigating 

---

Ok should be fixed, The recipe is indeed buggy (the remote has a merge commit for the regression based on 8.1.X or 9.0.X). Let's wait for the official announcement before reporting :)

swertz:
Twiki for EGMSmearing has been updated, no need to fetch giuseppe's repo anymore? https://twiki.cern.ch/twiki/bin/view/CMS/EGMSmearer#ECAL_scale_and_resolution_correc

blinkseb:
Let me try it :sushi:

blinkseb:
Much better indeed, works out-of-the-box. ~~I'll update the PR~~ Done.

blinkseb:
```

*** Discussion title: Egamma ID

Hi,

No, this is *not* the case because we updated the E-p regression method.
Right now we use a semi-parametric method with different inputs.

I will use the numbers you measured and release a consistent version
later in the day today.

Best,

-- Rafael.
```

Not sure what this really means, but we should probably wait a bit :worried:

swertz:
Indeed, what a mess...

OlivierBondu:
As Hugues was saying earlier: there is 'mess' in 'CMS' after all :smile:

blinkseb:
PR updated with latest smearing recipe from https://twiki.cern.ch/twiki/bin/viewauth/CMS/EGMRegression#Consistent_EGMSmearer

OlivierBondu:
many thanks !

PR #247, opened 2017-02-20
Electron SFs are binned in supercluster eta

blinkseb:
Title says everything :) No need to add an extra branch for the supercluster eta since electrons are accessible at analyzer level using the `products` vector of `edm::Ref<pat::Electron>`


OlivierBondu:
thanks !

PR #249, opened 2017-02-28
Ensure references are unique when creating a conditional branch

blinkseb:
Currently, all the conditional branches of the same type shares the same memory space (ie, the same static variable). This leads to branch content overwritten by other branches of the same type, and can be disastrous if one perform some check based on the content of the variable.

This PR removes this limitation by ensure every branches have its own memory space.

### DIFF
--- a/interface/Types.h
+++ b/interface/Types.h
@@ -15,8 +15,7 @@ namespace Framework {
         if (condition) {
             return tree[branch_name].write<T>();
         } else {
-            static T foo;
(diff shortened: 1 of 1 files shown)


blinkseb:
Jenkins is dead unfortunately, but this PR can be merged as it

OlivierBondu:
first time I ever click on the big red *Use your administrator privileges to merge this pull request.* :smile:

PR #250, opened 2017-03-07
Fix JER systematics

blinkseb:
JER systematics should be applied on a unsmeared jet collection, and not
on an already smeared jet collection as it's done currently.


OlivierBondu:
many thanks !

PR #251, opened 2017-03-14
Merge all option parsing into CmdLine

pieterdavid:
I noticed that the code to parse and validate command-line arguments and other options was a bit scattered (XXConfiguration.py, Framework/python/CmdLine.py and Framework/python/Framework.py).
This is a first attempt to put as much of that as possible in CmdLine - this has implications for Framework and the analysis configurations as well, and I may have missed important cases, so please have a look and give your opinion (not urgently). If we decide to go with this, we maybe should rename CmdLine into FrameworkOptions or something like that...
For an example of a configuration file using this, see https://github.com/pieterdavid/TTWAnalysis/blob/refactorCmdLine/test/TTWConfiguration.py (nothing spectacular, but the logic is a bit more explicit).
Last point: I took the opportunity to add the `Run2_2016` era as a valid option (even if that probably doesn't matter too much for our purpose at this point) - I'll move those changes to a separate PR and rebase once that's merged.

### DIFF
--- a/python/CmdLine.py
+++ b/python/CmdLine.py
@@ -9,54 +9,106 @@ class CmdLine(object):
         - 'runOnData': 1 if running on data, 0 otherwise
         - 'hltProcessName': the process name used when running the HLT
 
-    Used mainly by GridIn and crab to override user configuration of the framework with correct values
-    for a given dataset.
+    Can be customized from a config file (or better, subclass) using
+    changeDefaults(option=value, ...) and override(option=value, ...),
+    depending on whether the given value should be used is none is passed
+    on the command line, or if it should be used regardless of that.
+    Options can be added using options.register.
+    The registerOptions method (called from the constructor) can be
+    overridden by subclasses, e.g.
+    >>> from cp3_llbb.Framework.CmdLine import CmdLine
+    >>> class MyCmdLine(CmdLine):
+    >>>     def __init__(self):
+    >>>         super(MyCmdLine, self).__init__()
+    >>>     def registerOptions(self):
+    >>>         super(MyCmdLine, self).registerOptions()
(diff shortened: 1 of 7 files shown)


blinkseb:
Sorry for the very long delay! I really like the idea, it's always good to centralize things and this way it's more easier to add new options if ones are needed, thanks a lot! Looking around a bit at the code, I'm wondering if the `changeDefaults` method cannot be merged with the constructor of `CmdLine`? Maybe it'd be trickier to subclass if it's the case... Any idea?

I let the others review a bit, but for me it's good to go, I like the fact there's not too much to change at the analysis level.

Thanks :+1:

pieterdavid:
Thanks for having a look, I added also a "defaults" dictionary argument to the constructor, that makes it easier indeed (the `override` and `changeDefaults` method then need not be called in most cases, but I would tend to keep them for the benefit of subclasses - no strong opinion on that, though; test configs and docstrings are updated accordingly)

pieterdavid:
Given that it seems to be a reasonably quiet time production-wise, would now be a good time to merge this? (note: that will require a (small and straightforward) update of all analysis configs)
Any more thoughts/suggestions/objections?

blinkseb:
Thanks a lot again, looks very nice. Did you had a chance to test this for real with GridIn / crab? I want to be sure the flags are still correctly overridden when using crab.

pieterdavid:
I ran a production last week without problems... I checked the `PSetDump.py` files, and all tasks picked up the right global tag, i.e. the one in my config for MC because `runOnGrid.py` doesn't set that, and the explicitly specified ones for data (which differs from the one in my config for 2016H), so it looks like things are working fine.

blinkseb:
Then it's perfect! Thanks again!

PR #252, opened 2017-03-14
Add '2016' (Run2_2016) as a valid era

pieterdavid:


blinkseb:
I was actually thinking of removing the `era` altogether, since I'm not sure it's actually use anywhere in CMSSW, but it's still a nice addition, thanks! Is this PR mutually exclusive this #251 or can I merge this one and let you rebase the other?

pieterdavid:
I thought of rebasing the other one (this is simple and self-contained, and the other one isn't final yet anyway ;-) ). No strong opinion on keeping the option... if we pass the global tags explicitly, we may not need this indeed, but you never know in the future (some reconstruction options are era-versioned, if I remember correctly)

blinkseb:
It was indeed relevant when we were mixing 50ns and 25ns data. Does not harm too much to keep it, if we remove it, it'll probably start to become mandatory xD

So let's merge this one first :) Can you rebase it on top of `CMSSW_8_0_6p`?

blinkseb:
Perfect, looks like someone updated the official branches we are using for the recipe and the build fails because of that... :angry: I'll have a look later

pieterdavid:
Thanks! (it's the first time I needed `git push -f`, that looked a bit dangerous ;-) ). The compilation problem is with a graphics header (GL/glew.h) that's needed for fireworks... looks like somehow a dependency on that got introduced (maybe it's been moved to a new CMSSW version?)

blinkseb:
It's probably just one of the branch we merge now needs a higher version of CMSSW, which means we recompile Fireworks (...). But it'd be great if people stop updating their branch randomly, we now have almost zero way to reproduce our results since the recipe is dead. Maybe we should start having our own fork of CMSSW and merging recipes in it to ensure nothing is modified / deleted...

blinkseb:
I've created a new build image installing the OpenGL headers, let's see if it works!

please test

blinkseb:
please test again

pieterdavid:
I agree, we shouldn't have to follow moving targets. We can try to only merge things that refer to a specific commit (frozen branches or tags) - better than us forking cmssw, CMS could provide a `CMSSW_8_0_26_patch2-analysis` branch, with all these official recipes merged...

blinkseb:
Indeed that looks like the easiest and a robust way to go. The recipes is a bit less clear, but at least we are protected against crazy push months later!

This time should be the right one, Jenkins, please test

blinkseb:
We are close, please test

PR #254, opened 2017-04-10
Merge cms-met:METRecipe_8020 only up to the changes compatible with CMSSW_8_0_26_patch2

pieterdavid:
let's hope this fixes the failing jenkins build (probably best to specify or write down the specific commit for not explicitly versioned branches like this one, that should work as long as people don't force-push).
I don't see anything obviously incompatible in the more recent commits in that branch, it's probably through some newer/other packages these depend on


blinkseb:
Nice thanks! They updated the branch 3 days ago and pull some code from upstream (like https://github.com/cms-sw/cmssw/commit/8280ed793c21cab4f83eaee62a53a2b00e6c8841) which update fireworks, and since OpenGL headers are not installed on the build container, the build fails...

blinkseb:
Arf, can you rebase please?

PR #255, opened 2017-06-01
Add HZZ MVA electron ID

pieterdavid:
I am adding this one to my trees as an input for the latest training of the ttH lepton MVA (see https://twiki.cern.ch/twiki/bin/view/CMS/MultivariateElectronIdentificationRun2#HZZ_MVA_training_details_and_wor for details)


OlivierBondu:
Looks from the recipe that the input tag `egmGsfElectronIDs:mvaEleID-Spring16-HZZ-V1-wpLoose` should also be added to `python/ElectronsProducer.py` ?

pieterdavid:
Thanks for checking, added :-) . To also add the MVA classifier (and not only this WP) I'd need to do a bit more work... I'm directly reading that in my analyzer, but if there is interest I can add it

OlivierBondu:
and jenkins fails... because of course, an update of the reference trees is also needed, given you add a variable !

pieterdavid:
I should have known... last-minute before running to the train station, so I hope this works right away ;)

OlivierBondu:
Thanks !

Issue #257, opened 2017-09-13
Updates needed for 8.0.X / 2016 data analysis post-Moriond


OlivierBondu:
Copy-paste of https://github.com/cp3-llbb/Framework/issues/183

Just a small list of what we need to update for 8.0.X. Feel free to add more stuff!
- [x] Update CMSSW release
- [ ] Electrons
  - [x] Regression and smearing
  - [ ] RECO SFs 
  - [ ] effective areas
  - [x] ID
  - [x] ID SFs
  - [ ] trigger efficiencies and list in `data/triggers.xml` 
- [ ] Muons
  - [ ] tracking SF
  - [ ] SFs
  - [ ] trigger efficiencies and list in `data/triggers.xml`
  - [x] muon momentum scale (did not change)
- [ ] Jets
  - [ ] update jet IDs
  - [ ] update PU jet ID (see [the twiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/PileupJetID)) -- Postponed until 2016 rereco
  - [x] Add deepCSV (see https://twiki.cern.ch/twiki/bin/view/CMS/DeepFlavour)
  - [x] B-tagging SFs (we also need to split between light & c/b jets)
  - [ ] B-jet regression (see #178)
  - [x] MET
  - [x] MET significance
- [ ] PU profile

PR #258, opened 2017-09-19
update the cmssw release and MET recipe

OlivierBondu:
Getting started on https://github.com/cp3-llbb/Framework/issues/257

Let's try to keep the PR factorized on the different topics in order to properly check that changes are not breaking unexpected things...

First: updating the CMSSW release!

### DIFF
--- a/CMSSW.release
+++ b/CMSSW.release
@@ -1,1 +1,1 @@
-CMSSW_8_0_26_patch2
+CMSSW_8_0_30
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Common framework for all cp3-llbb analyses
 
 ## CMSSW release
 
-**CMSSW 8.0.26 patch 2**
+**CMSSW 8.0.30**
 
 ## First time setup instructions
 
@@ -18,8 +18,8 @@ source /nfs/soft/grid/ui_sl6/setup/grid-env.sh
 source /cvmfs/cms.cern.ch/cmsset_default.sh
 
 export SCRAM_ARCH=slc6_amd64_gcc530
-cmsrel CMSSW_8_0_26_patch2
-cd CMSSW_8_0_26_patch2/src
+cmsrel CMSSW_8_0_30
+cd CMSSW_8_0_30/src
 cmsenv
 
 git cms-init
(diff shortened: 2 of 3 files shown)


OlivierBondu:
please test

OlivierBondu:
Thanks @blinkseb to have restarted things up :)

It looks like the MET branch was updated wrt the commit jenkins was told to pick up, which is the reason why it compiled locally but not on jenkins... anyhow, looks like everything is fine again

Nota Bene: the MET recipe was not updated, but the old recipe was out of date, so updating MET recommendations is still on the todo list

pieterdavid (approved):
@OlivierBondu I don't really understand the comment about the MET commit (what can happen is that you need to merge another branch due to conflicting changes, but that should still be the same locally and in jenkins - as long as the original commit still exists upstream, which seems to be the case). Is there a twiki/hypernews ref for this recipe? Anyway, except documentation not much to discuss here, so fine with me.

OlivierBondu:
About the MET commit: what happened is that changing the release worked locally out of the box and not for jenkins. The reason is that
- the [readme instructions](https://github.com/cp3-llbb/Framework/blob/1a55ac3a5e008d5656c65b2b921d2fc00767a2ef/README.md#first-time-setup-instructions) are about `git cms-merge-topic cms-met:METRecipe_8020 -u`
- the [`bootstrap_jenkins.sh`](https://github.com/cp3-llbb/Framework/blob/a9cca5a32c179acb33c23370f63ad7e9daeb2a9f/bootstrap_jenkins.sh#L12) is about a specific commit of that branch, namely `git cms-merge-topic cms-met:83c1f5d9bfcbc669a0621cdd3c01893f47b268be -u` (a change you introduced btw :wink:)

what I did is just update the specific commit being picked up by jenkins

OlivierBondu:
but actually maybe we should revert the bootstrap script to `git cms-merge-topic cms-met:METRecipe_8020 -u`

blinkseb:
Please have a look here: https://twiki.cern.ch/twiki/bin/view/CMS/MissingETUncertaintyPrescription

Recipe looks updated

pieterdavid:
I missed that the hash change was in the second commit, thanks for explaining!
A hash or a branch name in the bootstrap script is a tradeoff between maintainability and reproducibility... I'd still go for the latter, but ideally they would make tags to make this easier ;-)

OlivierBondu:
I know, I just didn't want to update everything in one go... to factorize and keep track of changes on the ref trees, but indeed, maybe it's worth doing here as the current recipe is broken anyway

let me have a look then

pieterdavid:
Looks good, thanks!

PR #260, opened 2017-09-22
Update electron regression and smearing

OlivierBondu:
update instructions to latest twiki https://twiki.cern.ch/twiki/bin/view/CMS/EGMRegression

tracked at https://github.com/cp3-llbb/Framework/issues/257

_a priori_ there is no change in the reference trees from local tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,12 +34,16 @@ pushd KaMuCa
 git checkout 2ad38daae37a41a9c07f482e95f2455e3eb915b0
 popd
 
-# Electron smearing
-git cms-merge-topic rafaellopesdesa:EgammaAnalysis80_EGMSmearer_Moriond17_23Jan
+# Electron smearing consistent with regression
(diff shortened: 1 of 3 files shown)


pieterdavid:
You may need to add a `git fetch Moriond17_gainSwitch_unc` for ScalesSmearing, to make sure git finds the hash

update: please ignore, I misread the hashes

OlivierBondu:
For the record, if it is ever needed (you never know): the `git cms-merge-topic` hides a [`git cms-init`](https://github.com/cms-sw/cms-git-tools/blob/4c3fe993bd38bf7f248c840212c0583bc5ae062c/git-cms-merge-topic#L184), which is the reason why it was working before... (thanks to @pieterdavid for the help)

As I switched to pick up explicit commits instead of checkout of branches, it broke. Bref, it now finally compiles :smile:

OlivierBondu:
and success :tada:

pieterdavid (approved):

PR #261, opened 2017-09-22
adding DeepCSV and DeepCMVA, removing JetProbability

OlivierBondu:
tracked at https://github.com/cp3-llbb/Framework/issues/257

NB:
- I don't think any of us uses JetProbability, and we don't have SF for it, but it can easily be added back...
- no SF added at this time
- I expect the checks to fail, we should need to updated the ref trees... I'll do that on Monday


pieterdavid (approved):
Thanks, I didn't know these were available in the late 80 releases :-). The tests fail exactly as expected indeed, with missing/new branches

swertz:
FYI, CMVA is not really supported, and I think DeepCMVA isn't either. 

Also, the tagger tags are `pfDeepCSVJetTags:probX`, where X is `b`, `bb`, `c`, and `udsg`, and the sum of the first two is what actually defines the b-tagger.

pieterdavid:
Thanks @swertz . It turns out that we also misread "is available in the release" as "is saved in the miniAOD", so we need to add a bit of configuration to update the jets with these b-tags. I got something working yesterday evening, I will further test and clean it up on Monday.

OlivierBondu:
I pushed @pieterdavid changes which look fine to me :smile: 

Maybe @swertz should sign off on this one ?

pieterdavid:
Few more comments on my commits that @OlivierBondu just merged in his branch: there are a number of warnings printed, from the configuration
```
**************************************************************
b tagging needs to be run on uncorrected jets. Hence, the JECs
will first be undone for 'updatedPatJetsAK4PFchsNewJEC' and then applied to
'updatedPatJetsTransientCorrectedAK4PFchsNewJEC'.
**************************************************************
skipping pfDeepCSVJetTagsAK4PFchsNewJEC as it has already been loaded in the process
skipping pfDeepCSVJetTagsAK4PFchsNewJEC as it has already been loaded in the process
skipping pfDeepCSVJetTagsAK4PFchsNewJEC as it has already been loaded in the process
skipping pfDeepCSVJetTagsAK4PFchsNewJEC as it has already been loaded in the process
**************************************************************
```
which seem harmless, see [this hypernews thread](https://hypernews.cern.ch/HyperNews/CMS/get/btag/1399.html?inline=-1).

The other ones are these:
```
%MSG-w L3Absolute not found:  PATJetUpdater:updatedPatJetsAK4PFchsNewJEC  25-Sep-2017 14:09:33 CEST Run: 1 Event: 270939
L2L3Residual and L3Absolute are not part of the jetCorrFactors
of module patJetCorrFactorsAK4PFchsNewJEC. Jets will remain uncorrected.
%MSG
%MSG-w L3Absolute not found:  PATJetUpdater:updatedPatJetsAK8PFchsNewJEC  25-Sep-2017 14:09:41 CEST Run: 1 Event: 270940
L2L3Residual and L3Absolute are not part of the jetCorrFactors
of module patJetCorrFactorsAK8PFchsNewJEC. Jets will remain uncorrected.
%MSG
```
which look a bit worrisome to me, but no changes in the momenta etc., so I hope this is a red herring as well (it's a bit hard to understand what it really means - but I did not dig deeply into the code that produces it yet).

pieterdavid:
One last point: @alesaggio @OlivierBondu don't forget to update the ZA config (passing the list of DeepCSV outputs to redoJEC, like [here](https://github.com/cp3-llbb/Framework/pull/261/files#diff-6119f9d20a2562b419d64008e77c6d4dR12)) as well, otherwise all values will be -1000 ;-)

swertz:
Looks fine to me! About the strange logs: both messages are mentioned in the HN thread and are reported to be "normal"...

PR #262, opened 2017-09-25
Tests: exit a bit later (for a more detailed report)

pieterdavid:
- bin/testFrameworkOutput.cc: a different number of branches and deleted
  branches do still cause a failure, but no longer an early exit.
  This allows to see if anything else happened than adding/removing
  branches.
- test/run_tests.sh: even in case of failure other tests are still run


OlivierBondu:
thanks !

PR #263, opened 2017-09-27
Add an "initial project setup" script

pieterdavid:
Mostly for convenience (and to prove to @alesaggio that it's not that hard to change CMSSW versions ;-) ).
This also removes the duplication between the jenkins scripts and README.md (which now simply links to those), which should make it a bit less work to update the recipes.

### DIFF
--- a/setup_project_with_framework.sh
+++ b/setup_project_with_framework.sh
@@ -2,26 +2,17 @@
 
 startdir=$(pwd)
 
-cmssw_arch="slc6_amd64_gcc530" ## should be in sync with CMSSW.arch
-cmssw_version="CMSSW_8_0_30"   ## should be in sync with CMSSW.release
+tempdir=$(mktemp -d)
+fwkdir_tmp="${tempdir}/Framework"
+
+git clone -o upstream https://github.com/cp3-llbb/Framework "${fwkdir_tmp}"
+
+cmssw_arch=$(cat "${fwkdir_tmp}/CMSSW.arch")
+cmssw_version=$(cat "${fwkdir_tmp}/CMSSW.release")
 
 scramv1 -a "${cmssw_arch}" project CMSSW "${cmssw_version}"
 cd "${cmssw_version}"
 cmsenv
-
-git clone -o upstream https://github.com/cp3-llbb/Framework
-fwkdir_tmp="${CMSSW_BASE}/Framework"
-
-upstream_cmssw_arch=$(cat "Framework/CMSSW.arch")
-if [ "${upstream_cmssw_arch}" != "${cmssw_arch}" ]; then
-  echo "Platform mismatch with CMSSW.arch: ${upstream_cmssw_arch}, while set up with ${cmssw_arch}"
-  exit 1
-fi
-upstream_cmssw_version=$(cat "Framework/CMSSW.release")
-if [ "${upstream_cmssw_version}" != "${cmssw_version}" ]; then
-  echo "CMSSW version mismatch with CMSSW.release: ${upstream_cmssw_version}, while set up with ${cmssw_version}"
-  exit 1
-fi
 
 ## apply recipes
 cd "${CMSSW_BASE}/src"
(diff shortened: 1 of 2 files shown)


pieterdavid:
Sorry for pushing more commits after creating the PR, each of one which triggered a jenkins build... I cancelled the first two (jenkins doesn't read README.md, so there shouldn't be any risk).

setup_project_with_framework.sh
```diff
@@ -0,0 +1,47 @@
+#! /bin/sh
+
+startdir=$(pwd)
+
+cmssw_arch="slc6_amd64_gcc530" ## should be in sync with CMSSW.arch
+cmssw_version="CMSSW_8_0_30"   ## should be in sync with CMSSW.release
```
OlivierBondu: why not read it from `CMSSW.arch` and `CMSSW.release` files then ? This would again avoid having to do it twice ?
I see your warnings below but I am not sure to see the use case, is it for whenever we would want to test an unstable recipe before it is merged ?

setup_project_with_framework.sh
```diff
@@ -0,0 +1,47 @@
...
+  exit 1
+fi
+
+## apply recipes
+cd "${CMSSW_BASE}/src"
+. "${fwkdir_tmp}/bootstrap_jenkins.sh"
+## move Framework to the right place
+llbbdir="${CMSSW_BASE}/src/cp3_llbb"
+mkdir -p "${llbbdir}"
+mv "${fwkdir_tmp}" "${llbbdir}/"
```
OlivierBondu: I'm not sure I like moving repos after checking out... basically the only thing that prevents doing the clone directly in the right place is the `git cms-init` if I am not mistaken, maybe adding one in this file should not hurt calling the `bootstrap_jenkins.sh` after ?

setup_project_with_framework.sh
```diff
@@ -0,0 +1,47 @@
...
+## compile
+scram b -j4
```
OlivierBondu: if this is supposed to be on `ingrid-ui2`, using only 4 cores is a bit of underkill, we don't compile so often this would disturb all the users ?

OlivierBondu (changes requested):
Thanks ! looks fine, I have some questions because I am not sure to understand everything, but nothing critical :)

setup_project_with_framework.sh
pieterdavid: That's easy to do, but I wanted to make the "initial setup" as light as possible, and now this is done before the Framework is cloned. But there is a solution to that: I can clone it to a temporary directory, then read them from there, do everything up to the bootstrap, and then move it into place. I will update this

setup_project_with_framework.sh
pieterdavid: it's more tricky: `git cms-init` needs the `${CMSSW_BASE}/src` directory to be empty, so as long as one is there in `bootstrap_jenkins.sh` you cannot have Framework in the right place until after that is called.

setup_project_with_framework.sh
pieterdavid: I'm not sure you gain much with >4... what do you usually use?

pieterdavid: -j7 now :)

setup_project_with_framework.sh
pieterdavid: Having this for unstable recipes would be nice, I'll see if it is easy to implement by taking a single argument

setup_project_with_framework.sh
pieterdavid: moving a git repository should be fine, see https://stackoverflow.com/questions/17568543/git-add-doesnt-work/17747571#17747571 for details (we don't even use submodules)

OlivierBondu: ok, fine then :) thanks !

pieterdavid:
I integrated the nice features from @blinkseb 's [frameworkBootstrap.sh](https://gist.github.com/blinkseb/ef079e273a5b28ca95348f62c4f6f2c5#file-frameworkBootstrap-sh).

The main differences with respect to that are:
- changed branch to a named option, and added `--pr`
- kept the "checkout to temporary then move" strategy, which makes sure that `CMSSW.version`, `CMSSW.arch` and `framework_bootstraps.sh` correspond to the version that is being checked out
- changed `exit` to `return` (otherwise it exits the calling shell if it was sourced).

OlivierBondu (approved):

OlivierBondu:
Thanks !

PR #265, opened 2017-09-28
Checks on the order of Framework config method calls

pieterdavid:
implemented using a decorator class with "before", "after" and
"performs" tag lists, which will enforce none of "before" flags has been
set yet, all of the "after" flags have been set, and set the "performs"
flags to True.

There are some limitations to the scaling of this, but I wanted to keep it reasonably light (both the implementation and amount of annotations to add) - feel free to try and break it from your config by doing things in the wrong order ;-) . It should enforce that adding/removing analyzers&producers can only be done before any of the corrections, as well as the orderings (JEC only before JER, electron smearing needs regression) that were already checked for.

### DIFF
--- a/python/Framework.py
+++ b/python/Framework.py
@@ -90,7 +90,7 @@ class Framework(object):
         process.maxEvents = cms.untracked.PSet(input = cms.untracked.int32(20))
         process.source = cms.Source("PoolSource")
 
-        self.configureFramework_()
+        self._configureFramework()
 
     @dep(before="create", performs="create", fallback=lambda self : self.process)
     def create(self):
@@ -527,8 +527,7 @@ class Framework(object):
 
             self.process.electronMVAValueMapProducer.srcMiniAOD = self.__miniaod_electron_collection
 
-    @dep(before="configure", performs="configure")
-    def configureFramework_(self):
+    def _configureFramework(self):
 
         from cp3_llbb.Framework import EventProducer
         from cp3_llbb.Framework import GenParticlesProducer
--- a/python/Framework.py
+++ b/python/Framework.py
@@ -5,6 +5,7 @@ from FWCore.ParameterSet.SequenceTypes import _SequenceCollection
 
 from Configuration.StandardSequences.Eras import eras
 
+from cp3_llbb.Framework.deptracker import deptracker as dep
 from cp3_llbb.Framework.Tools import change_process_name, change_input_tags_and_strings, StdStreamSilenter
 from cp3_llbb.Framework import Tools
 import cp3_llbb.Framework.Systematics as Systematics
@@ -12,11 +13,8 @@ import cp3_llbb.Framework.Systematics as Systematics
 class Framework(object):
 
     def __init__(self, options, verbose=True):
-        self.__created = False
         self.__systematics = []
         self.__systematicsOptions = {}
-        self.__jec_done = False
-        self.__jer_done = False
 
         # Default configuration
         self.__miniaod_jet_collection = 'slimmedJets'
@@ -29,10 +27,6 @@ class Framework(object):
 
         self.__jer_resolution_file = None
         self.__jer_scalefactor_file = None
-
-        self.__electron_regression_done = False
-        self.__electron_smearing_done = False
-        self.__muon_correction_done = False
 
         self.__kamuca_tag = 'DATA_80X_13TeV' if options.runOnData else 'MC_80X_13TeV'
         self.__rochester_input = 'cp3_llbb/Framework/data/Rochester/2016.v3/config.txt'
@@ -98,13 +92,11 @@ class Framework(object):
 
         self.configureFramework_()
 
+    @dep(before="create", performs="create", fallback=lambda self : self.process)
     def create(self):
         """
         Terminate the process configuration and return the final CMSSW process
         """
-
-        if self.__created:
-            return self.process
 
         self.configureElectronId_()
 
@@ -163,15 +155,13 @@ class Framework(object):
         self.process.framework.analyzers_scheduling = cms.untracked.vstring(self.analyzers)
         self.process.framework.producers_scheduling = cms.untracked.vstring(self.producers)
 
-        self.__created = True
         return self.process
 
+    @dep(before=("create", "correction"))
     def addAnalyzer(self, name, configuration, index=None):
         """
         Add an analyzer in the framework configuration with a given name and configuration
         """
-
-        self.ensureNotCreated()
 
         if name in self.analyzers:
             raise Exception('An analyzer named %r is already added to the configuration' % name)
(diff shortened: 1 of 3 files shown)


pieterdavid:
oops, I should have checked the test... been a bit too paranoid, I'm adding an optional argument to skip some checks

python/Framework.py
```diff
@@ -98,14 +92,12 @@ def __init__(self, options, verbose=True):
 
         self.configureFramework_()
 
+    @dep(before="create", performs="create", fallback=lambda self : self.process)
```
OlivierBondu: what's the fallback for?

python/Framework.py
```diff
@@ -328,15 +307,12 @@ def redoJEC(self, JECDatabase=None, addBtagDiscriminators=None):
 
         print("")
 
-        self.__jec_done = True
-
+    @dep(before=("create", "jer"), performs=("correction", "jer"))
     def smearJets(self, resolutionFile=None, scaleFactorFile=None):
```
OlivierBondu, line 311: would it be possible to keep in sync what the function does (smear jets) and the name of the action performed (jer) ? This may ease debugging :smile:

python/Framework.py
```diff
@@ -396,8 +372,7 @@ def smearJets(self, resolutionFile=None, scaleFactorFile=None):
         if self.verbose:
             print("New jets and MET collections: %r and %r" % (self.__miniaod_jet_collection, self.__miniaod_met_collection))
 
-        self.__jer_done = True
-
+    @dep(before=("create", "muonScale"), performs=("correction", "muonScale"))
     def applyMuonCorrection(self, type):
```
OlivierBondu, line 376: Same here, muon scale and muon correction are not necessarily the exact same thing

python/Framework.py
```diff
@@ -440,16 +413,13 @@ def applyMuonCorrection(self, type):
         if self.verbose:
             print("New muons collection: %r" % (self.__miniaod_muon_collection))
 
-        self.__muon_correction_done = True
-
+    @dep(before=("create", "eleReg"), performs=("correction", "eleReg"))
```
OlivierBondu: as the previous flag was 'muonScale' I would prefer here "electronRegression", longer, but also more explicit

python/Framework.py
```diff
@@ -473,23 +443,17 @@ def applyElectronRegression(self):
         if self.verbose:
             print("New electrons collection: %r" % (self.__miniaod_electron_collection))
 
-        self.__electron_regression_done = True
-
+    @dep(before=("create", "eleSmear"), after="eleReg", performs=("eleSmear", "correction"))
```
OlivierBondu: same here --> "electronSmearing"

python/Framework.py
```diff
@@ -571,10 +527,9 @@ def configureElectronId_(self):
 
             self.process.electronMVAValueMapProducer.srcMiniAOD = self.__miniaod_electron_collection
 
+    @dep(before="configure", performs="configure")
     def configureFramework_(self):
```
OlivierBondu: this configuration do not seem to be compulsory in the current status ?

python/deptracker.py
```diff
@@ -0,0 +1,81 @@
...
+    >>> @deptracker(before=("create", "eleSmear"), after="eleReg", performs=("eleSmear", "correction"))
+    >>> def applyElectronSmearing(self):
+    >>>     ...
+
+    which will, when calling this method, check that no member method with `@deptracker(..., performs=("create"))`
+    nor `@deptracker(..., performs=("eleSmear"))` has been called yet, and
+    that a method with `@deptracker(..., performs=("eleReg", ...))` has been called.
+    Furthermore, this will mark "eleSmear" and "correction" as "done", such that no other method
+    with either of these in "before" can be called after (for "eleSmear": including this method;
+    putting the same tag in "before" and "after" ensures that a method is called no more than once).
```
OlivierBondu: same here in documentation, I would prefer explicit naming of the "tags" rather than abbreviations

OlivierBondu (changes requested):
Thanks! looks good to me at first look :smile: 

Now I'll try testing this :stuck_out_tongue:

python/Framework.py
pieterdavid: It used to return the process if `create` had been called before... I'm not sure whether that's used anywhere (it would indeed be nicer not to need this, we could also make `process` a property that triggers `create` if not already done so, and use `_process` instead internally, which one do you prefer?)

python/Framework.py
pieterdavid: Sure, I used the `__jer_done` to get a name, but this is easy to change. On the other hand, the "action" needs not to say exactly what the function does (it's only used for ordering things logically). Not an expert here, "smearJets" also sounds good to me

OlivierBondu:
testing looks fine :smile: 

many thanks!

python/Framework.py
pieterdavid: Yes... but "correction" can mean pretty much anything, and this is all momentum calibration, therefore I'd call this a "scale correction" (in the hypothetical case where we need to do some ID I think this is clearer), but that's a personal opinion, I don't feel strongly about this.

python/Framework.py
pieterdavid: ok, I'll change this one

python/Framework.py
pieterdavid: yes

python/Framework.py
pieterdavid: It prevents this method being called again (not from the constructor), because it should only be called once. We could also leave this out and make it `_configurFramework`, following the python convention that such a method is private then.

python/deptracker.py
pieterdavid: ok

python/Framework.py
OlivierBondu: boh ok, for this one we can skip, I agree !

python/Framework.py
OlivierBondu: mmm, now this rings a (distant) bell, but this was not present before ?

python/Framework.py
OlivierBondu: ok, let's keep jer :)

python/Framework.py
OlivierBondu: but this is called only once when the constructor is called no ? I find the use of the decorator confusing here, because this is not something you are supposed to be playing with at analysis config level anyway ?

python/Framework.py
pieterdavid: it used to start with `if self.__created: return self.process`, but indeed this feature seems not to be used, as far as I can see, so we can remove it, if you agree

python/Framework.py
pieterdavid: Yes, changed to `_configureFramework` and removed the decorator

python/Framework.py
OlivierBondu: agreed then!

python/Framework.py
OlivierBondu: thanks!

OlivierBondu (approved):

OlivierBondu:
Thanks! let's wait for jenkins to compile :)

pieterdavid:
I wanted to cancel the jenkins job for the first of the two commits I just pushed, and saw that it's stuck... I tried following Sébastien's instructions, but I'm not an admin, so I can't restart the container

OlivierBondu:
please test

OlivierBondu:
please test

PR #266, opened 2017-10-23
Allow building BinnedValues(JSONParser) as a standalone library

pieterdavid:
Changing to CMSSW_8_0_30, I noticed that I had not upstreamed these changes yet... the reason to build a library from BinnedValues and BinnedValuesJSONParser that doesn't depend on CMSSW is that I can then more easily link my plotter to this one instead of copying the code :-) (I apply SFs at that stage). I know ifdef's are not nice... but I hope these are few and hidden enough to be acceptable ;-) (they are only in the cc files).
I also removed some includes from ScaleFactors.h and BTaggingScaleFactors.h, but those should not be affected - it's only because I saw the same list in three places, so probably they've been copied around before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.lo
 *.o
 *.obj
+*.d
 
 # Precompiled Headers
 *.gch
--- a/Makefile.libBinnedValues
+++ b/Makefile.libBinnedValues
@@ -1,0 +1,56 @@
+## adapted from plotIt/Makefile
+
+ObjSuf = o
+SrcSuf = cc
+DllSuf = so
+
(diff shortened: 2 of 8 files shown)


pieterdavid:
test this please

pieterdavid:
@cp3-llbb-bot test this please

pieterdavid:
@cp3-llbb-bot test this please

OlivierBondu:
Thanks ! The build should be triggered automatically... but in any case, the correct command should be `please test`, the README is not up to date... we fixed this a while ago to be in sync with the CMSSW repo :)

PR #273, opened 2018-02-15
adding electron track expectedMissingInnerHits

OlivierBondu:


src/ElectronsProducer.cc
```diff
@@ -56,6 +56,8 @@ void ElectronsProducer::produce(edm::Event& event, const edm::EventSetup& eventS
         ecalPFClusterIso.push_back(electron.ecalPFClusterIso());
         hcalPFClusterIso.push_back(electron.hcalPFClusterIso());
         trackIso.push_back(electron.trackIso());
+        // see https://github.com/latinos/LatinoTrees/blob/cc89e18d1a956f633cf59a7143f24491716d1047/DataFormats/src/SkimEvent.cc#L5831
+        expectedMissingInnerHits.push_back(electron.gsfTrack()->hitPattern().numberOfHits(reco::HitPattern::MISSING_OUTER_HITS));
```
pieterdavid: It's a bit confusing to fill the number of missing outer hits in the expectedMissingInnerHits branch... or am I missing something?

OlivierBondu: I copy-pasted the wrong line :sweat_smile: thanks for checking !!

OlivierBondu:
(finally) with updated ref trees, it was the only failure in the previous log

PR #274, opened 2018-02-15
Workaround for 'git cms-merge-topic <hash>'

pieterdavid:
I copied the few relevant lines from `git-cms-merge-topic`, and assumed we always take a commit. The tricky part is related to sparse-checkout (figuring out the merge base, and using that to update the sparse-checkout config, but the hard part is done by cms-sparse-checkout).
I also took the freedom to reduce the number of `git cms-checkdeps` calls to one, that saves a bit of printout.

### DIFF
--- a/bootstrap_jenkins.sh
+++ b/bootstrap_jenkins.sh
@@ -19,7 +19,7 @@ function cms-merge-commit()
 # CMSSW env is already configured
 # Current working dir is $CMSSW_BASE/src
 
-git cms-init
+git cms-init --upstream-only
 
 echo "---> Merging recipes..."
 cms-merge-commit cms-egamma EGM_gain_v1 f2ae5ef247d2544bdccd0460586b468eb35631c5
--- a/bootstrap_jenkins.sh
+++ b/bootstrap_jenkins.sh
@@ -1,22 +1,39 @@
 #! /bin/sh
 
+function cms-merge-commit()
+{
+  local github_user="${1}"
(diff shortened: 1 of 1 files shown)


pieterdavid:
That was quick... too bad jenkins isn't using zsh instead of bash :(

blinkseb:
Try changing the shebang of the script, it may help :)

pieterdavid:
Thanks for the suggestion, but it sources the bootstrap script (with `. .../bootstrap_jenkins.sh`), so that doesn't help :-(, I'll check the jenkins config

pieterdavid:
@cp3-llbb-bot : please test

OlivierBondu:
Let's wait for @pieterdavid to have his look at the `cms-merge-topic` history before merging ?

pieterdavid:
The log looks fine now :-) . For reference: this broke with [this change](https://github.com/cms-sw/cms-git-tools/commit/8748af1f498ddd4537c1dc3690dd091b81fe5bc7#diff-ad03cccb399865882a3703e23f8e105fR127) (12 December 2017) in git-cms-merge-topic, so release areas from before that should be fine.
I pushed another commit to use `git cms-init --upstream-only` which should save quite some time and printout (because it doesn't need to get all the old tags), let's wait for the tests and then it should be fine to merge

pieterdavid:
@cp3-llbb-bot please test

OlivierBondu:
Thanks !

PR #276, opened 2018-02-16
Run the diff script also in generate_reference_trees.sh

pieterdavid:
It is supposed to "fail" (show differences), but will show what changes are about to be pushed.
This should save a step (and prevent common errors - maybe we should even add `scram b`?)


OlivierBondu:
excellent idea :)

as for the `scram b`, I'm guessing crab in no way knows about the hash of the packages it compiled last ? This would be the ideal: suggest to build only if there are changes (bref, `make` as it should be)

(also, looks like I tagged a bit too soon, sorry about that, but well it shouldn't matter for running a production...)

OlivierBondu:
Thanks a lot!!

PR #277, opened 2018-02-26
Added missing brackets

mdelcourt:
Added seemingly missing brackets

```
EventProducer.cc:276:13: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
             if (m_pu_reweighter.get())
             ^~
EventProducer.cc:278:17: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the 'if'
                 pu_weight_up = m_pu_reweighter_up->getWeight(true_interactions);
```


pieterdavid:
Thanks!

PR #278, opened 2018-03-23
Updated framework to run on phase2 data

mdelcourt:
- Adapted to run on CMSSW 9_3_6
- Added recipes from HGCAL TDR

### DIFF
--- a/bootstrap_jenkins.sh
+++ b/bootstrap_jenkins.sh
@@ -21,19 +21,8 @@ function cms-merge-commit()
 
 git cms-init --upstream-only
 
-echo "---> Merging recipes..."
-cms-merge-commit cms-egamma EGM_gain_v1 f2ae5ef247d2544bdccd0460586b468eb35631c5
-cms-merge-commit rafaellopesdesa RegressionCheckNegEnergy 3aafeff0371a1d1eb3db9d95ef50c1a66da25690
-cms-merge-commit cms-met METRecipe_8020_for80Xintegration 92f73cd3d16a9529585865a365de271e0535b68d
-
 echo "---> Running git cms-checkdeps"
 git cms-checkdeps -a
-
-echo "---> Adding KaMuCa"
-git clone -o upstream https://github.com/bachtis/analysis.git -b KaMuCa_V4 KaMuCa 
-pushd KaMuCa
-git checkout 2ad38daae37a41a9c07f482e95f2455e3eb915b0
-popd
 
 echo "---> Adding TreeWrapper"
 git clone -o upstream https://github.com/blinkseb/TreeWrapper.git cp3_llbb/TreeWrapper
--- a/bootstrap_jenkins.sh
+++ b/bootstrap_jenkins.sh
@@ -26,3 +26,10 @@ git cms-checkdeps -a
 
 echo "---> Adding TreeWrapper"
 git clone -o upstream https://github.com/blinkseb/TreeWrapper.git cp3_llbb/TreeWrapper
+
+echo "---> Adding phase2 EGamma"
+git cms-merge-topic -u nsmith-:EgammaFromMultiCl_932v2
+mkdir -p RecoEgamma && pushd RecoEgamma
+git clone -b integrated git@github.com:nsmith-/Phase2InterimID.git
+popd
+
--- a/bootstrap_jenkins.sh
+++ b/bootstrap_jenkins.sh
@@ -21,9 +21,6 @@ function cms-merge-commit()
 
 git cms-init --upstream-only
 
-echo "---> Running git cms-checkdeps"
-git cms-checkdeps -a
-
 echo "---> Adding TreeWrapper"
 git clone -o upstream https://github.com/blinkseb/TreeWrapper.git cp3_llbb/TreeWrapper
 
@@ -33,3 +30,5 @@ mkdir -p RecoEgamma && pushd RecoEgamma
 git clone -b integrated git@github.com:nsmith-/Phase2InterimID.git
 popd
 
+echo "---> Running git cms-checkdeps"
+git cms-checkdeps -a
--- a/interface/CorrectedMuonProducerT.h
+++ b/interface/CorrectedMuonProducerT.h
@@ -23,38 +23,12 @@
 #include "FWCore/ParameterSet/interface/ParameterSetDescription.h"
 
 #include <cp3_llbb/Framework/interface/Rochester.h>
-#include <KaMuCa/Calibration/interface/KalmanMuonCalibrator.h>
 
 #include <memory>
 #include <random>
 #include <boost/filesystem.hpp>
 
 namespace cp3 {
-
-    class KaMuCaCorrector {
-        public:
-            explicit KaMuCaCorrector(const edm::ParameterSet& cfg) {
-                std::string tag = cfg.getParameter<std::string>("tag");
-                corrector.reset(new KalmanMuonCalibrator(tag));
-            }
-
-            template<typename T>
-            T correct(const edm::Event& event, const T& muon) {
-                auto corrected_pt = corrector->getCorrectedPt(muon.pt(), muon.eta(), muon.phi(), muon.charge());
-                if (! event.isRealData())
-                    corrected_pt = corrector->smear(corrected_pt, muon.eta());
-
-                double ratio = corrected_pt / muon.pt();
-
-                T corrected_muon = muon;
-                corrected_muon.setP4(corrected_muon.p4() * ratio);
-
-                return corrected_muon;
-            }
-
-        private:
-            std::unique_ptr<KalmanMuonCalibrator> corrector;
-    };
 
     class RochesterCorrector {
         public:
--- a/python/ElectronsProducer.py
+++ b/python/ElectronsProducer.py
@@ -5,38 +5,31 @@ default_configuration = cms.PSet(
         prefix = cms.string('electron_'),
         enable = cms.bool(True),
         parameters = cms.PSet(
-            src = cms.untracked.InputTag('slimmedElectrons'),
+            src = cms.untracked.InputTag('phase2Electrons'),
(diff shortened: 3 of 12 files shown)


interface/CorrectedMuonProducerT.h
```diff
@@ -23,39 +23,13 @@
 #include "FWCore/ParameterSet/interface/ParameterSetDescription.h"
 
 #include <cp3_llbb/Framework/interface/Rochester.h>
-#include <KaMuCa/Calibration/interface/KalmanMuonCalibrator.h>
```
pieterdavid, line 26: Does it give problems, or do you simply not need KaMuCa? If it's the latter you could also just leave it in without using it (in the recipe and here), the C++ doesn't diverge so much then

python/ElectronsProducer.py
```diff
@@ -5,38 +5,31 @@
...
-                'egmGsfElectronIDs:mvaEleID-Spring16-GeneralPurpose-V1-wp90',
-                'egmGsfElectronIDs:mvaEleID-Spring16-GeneralPurpose-V1-wp80',
-                'egmGsfElectronIDs:mvaEleID-Spring16-HZZ-V1-wpLoose'
+#                'No ids defined for phase2'
                 ),
-            mva_id = cms.untracked.PSet(
-                values = cms.untracked.InputTag('electronMVAValueMapProducer:ElectronMVAEstimatorRun2Spring16GeneralPurposeV1Values'),
-                categories = cms.untracked.InputTag('electronMVAValueMapProducer:ElectronMVAEstimatorRun2Spring16GeneralPurposeV1Categories')
-            ),
+#            mva_id = cms.untracked.PSet(
```
pieterdavid: Mostly a matter of style, but there's no need to keep these around commented, it only adds confusion (for the mva_id it could be argued that they show the structure, but the whole list of scalefactors is definitely not needed)

python/Framework.py
```diff
@@ -77,10 +77,10 @@ def __init__(self, options, verbose=True):
                 input_files = cms.vstring()
                 )
 
-        process.load('Configuration.StandardSequences.FrontierConditions_GlobalTag_condDBv2_cff')
-        process.load('Configuration.StandardSequences.GeometryRecoDB_cff')
-        process.load('Configuration.StandardSequences.MagneticField_38T_cff')
         process.load('FWCore.MessageLogger.MessageLogger_cfi')
+        process.load('Configuration.StandardSequences.FrontierConditions_GlobalTag_cff')
+        process.load('Configuration.StandardSequences.MagneticField_cff')
+        process.load('Configuration.Geometry.GeometryExtended2023D17Reco_cff')
```
pieterdavid, line 83: Forgive my ignorance, but can't you get the geometry from the database (using the global tag)?

src/ElectronsProducer.cc
```diff
@@ -65,6 +65,32 @@ void ElectronsProducer::produce(edm::Event& event, const edm::EventSetup& eventS
...
+
+        bool passesMedium = false;
+        passesMedium += !electron.isEB() && electron.pt() > 20 && electron.userFloat("mvaValue") > 0.591;
+        passesMedium += !electron.isEB() && electron.pt() > 10 && electron.pt() <=20 && electron.userFloat("mvaValue") > -0.777;
+        passesMedium += electron.isEB() && electron.pt() > 20 && electron.userFloat("mvaValue") > 0.723;
+        passesMedium += electron.isEB() && electron.pt() > 10 && electron.pt() <=20 && electron.userFloat("mvaValue") > 0.855;
+        phase2_mva_id_medium.push_back(passesMedium);
+
+        bool passesTight = false;
+        passesMedium += !electron.isEB() && electron.pt() > 20 && electron.userFloat("mvaValue") > 0.983;
```
OlivierBondu: the name of the boolean is wrong here :)

python/Framework.py
```diff
@@ -509,23 +509,8 @@ def doSystematics(self, systematics, **kwargs):
...
-                    'RecoEgamma.ElectronIdentification.Identification.mvaElectronID_Spring16_GeneralPurpose_V1_cff',
-                    'RecoEgamma.ElectronIdentification.Identification.mvaElectronID_Spring16_HZZ_V1_cff'
-                    ]
-
-            for mod in id_modules:
-                setupAllVIDIdsInModule(self.process, mod, setupVIDElectronSelection)
-
-            self.process.electronMVAValueMapProducer.srcMiniAOD = self.__miniaod_electron_collection
+          self.process.load("RecoEgamma.Phase2InterimID.phase2EgammaPAT_cff")
+          self.path += cms.Sequence(self.process.phase2Egamma)
```
pieterdavid, line 513: There's nothing wrong with adding the module to the path explicitly, but we use "on-demand scheduling" everywhere for the framework (this changed in 91 or 92, you now need to add "tasks"), could you try and see if `self.path.associate(cms.Task(self.process.phase2Egamma))` also works?

OlivierBondu (changes requested):
thanks ! One typo, otherwise it looks good to me

src/ElectronsProducer.cc
mdelcourt: Thanks!

OlivierBondu:
Question: we should also update what jenkins needs to check here (new files, no data), should this be a separate PR ?

OlivierBondu (approved):

src/ElectronsProducer.cc
```diff
@@ -65,6 +65,32 @@ void ElectronsProducer::produce(edm::Event& event, const edm::EventSetup& eventS
...
+        passesMedium += !electron.isEB() && electron.pt() > 10 && electron.pt() <=20 && electron.userFloat("mvaValue") > 0.969;
+        passesMedium += electron.isEB() && electron.pt() > 20 && electron.userFloat("mvaValue") > 0.988;
+        passesMedium += electron.isEB() && electron.pt() > 10 && electron.pt() <=20 && electron.userFloat("mvaValue") > 0.986;
+        phase2_mva_id_tight.push_back(passesTight);
+
+
```
pieterdavid: Readability suggestion:
```C++
const float mva_value = electron.userFloat("mvaValue");
bool passesLoose{false}, passesMedium{false}, passesTight{false};
if ( electron.isEB() ) {
  if ( electron.pt() > 20. ) {
    passesLoose  = mvaValue > -0.797;
    passesMedium = mvaValue >  0.723;
    passesTight  = mvaValue >  0.988;
  } else if ( electron.pt() > 10. ) {
    // ...
  }
} else {
  if ( electron.pt() > 20. ) {
    // ...
  } else if ( electron.pt() > 10. ) {
    // ...
  }
}
phase2_mva_id_loose.push_back(passesLoose);
phase2_mva_id_medium.push_back(passesMedium);
phase2_mva_id_tight.push_back(passesTight);
```
it is a matter of taste, up to some point... but less repetition usually means fewer opportunities for typos to cause bugs

bootstrap_jenkins.sh
```diff
@@ -21,19 +21,15 @@ function cms-merge-commit()
...
-git clone -o upstream https://github.com/bachtis/analysis.git -b KaMuCa_V4 KaMuCa 
-pushd KaMuCa
-git checkout 2ad38daae37a41a9c07f482e95f2455e3eb915b0
-popd
-
 echo "---> Adding TreeWrapper"
 git clone -o upstream https://github.com/blinkseb/TreeWrapper.git cp3_llbb/TreeWrapper
+
+echo "---> Adding phase2 EGamma"
+git cms-merge-topic -u nsmith-:EgammaFromMultiCl_932v2
```
pieterdavid: This should go above `git cms-checkdeps -a` (you never know). We merge a specific commit to make sure things don't change when some adds a commit to a branch, but you probably don't need / want that level of paranoia in this case

pieterdavid (changes requested):
Thanks, looks good! I added a few questions and suggestions, but hardly anything that needs to be changed.

interface/CorrectedMuonProducerT.h
mdelcourt: It wouldn't compile under CMSSW_9_2_6p2, that's why I removed it.

pieterdavid:
@OlivierBondu everything related to the tests is also in git, so whatever works for @mdelcourt (e.g. a Delphes and a full sim sample as tests, the one with JEC from db may not make much sense here), I would say

pieterdavid:
The tests are failing with (mc)
```
No module named regressionWeights_cfi
```
and (mc with DB, data)
```
No module named Phase2InterimID.phase2EgammaPAT_cff
```
so they need at least the same changes as the HH config (no corrections and systematics), and then also new reference files

python/Framework.py
mdelcourt: When I tried without the GeometryExtended2023D17Reco_cff, I got a segmentation fault in the package computing HGC electron identification.

interface/CorrectedMuonProducerT.h
pieterdavid: This depends on adding the KaMuCa package in the bootstrap script, but I agree we shouldn't be patching it ;-)

python/Framework.py
mdelcourt: ----- Begin Fatal Exception 23-Mar-2018 16:33:50 CET-----------------------
An exception of category 'ConfigFileReadError' occurred while
   [0] Processing the python configuration file named HHConfiguration.py
Exception Message:
python encountered the error: <type 'exceptions.RuntimeError'>
Adding an entry of type 'Sequence'to a Task.
It is illegal to add this type to a Task.
----- End Fatal Exception -------------------------------------------------

pieterdavid: Let's leave it like this for now then (this will come back when you add the systematics)

pieterdavid (approved):

pieterdavid:
@mdelcourt should I merge this, or do you want to do something with the tests in this PR still?

pieterdavid:
In the end the framework is not used for the Phase2 studies (since the Delphes samples are not MiniAOD), but after discussing with @mdelcourt I am merging this PR anyway, in case the code changes are useful at some point (they go into the separate `phase2` branch, so should not cause inconvenience for anyone).

PR #280, opened 2018-08-08
Updates for CMSSW_9_4_X

pieterdavid:
I checked out the Framework in CMSSW_9_4_9, removed everything that said 80X or 'Moriond17' etc., and tried to make things run... so there will be differences, and I may have been too enthusiastically removing things in some places - my main aim was to provide a "clean" starting point, but feel free to point out if something doesn't make sense.

Anticipating a few questions:
- on-demand scheduling requires a `cms.Task` to be associated with the process or path for each `edm::EDProducer`. This required quite a bit of trial-and-error with redoJEC etc., but the tests should run now
- I turned off EGM regression and smearing for now, because the way this works changed (new recipes at https://twiki.cern.ch/twiki/bin/view/CMS/EgammaMiniAODV2 and https://twiki.cern.ch/twiki/bin/view/CMS/EgammaPostRecoRecipes) and I haven't implemented the new recipes yet.


README.md
```diff
@@ -18,12 +23,14 @@ Common framework for all cp3-llbb analyses
 source /nfs/soft/grid/ui_sl6/setup/grid-env.sh
 source /cvmfs/cms.cern.ch/cmsset_default.sh
 
-wget https://raw.githubusercontent.com/cp3-llbb/Framework/CMSSW_8_0_6p/setup_project_with_framework.sh
-source setup_project_with_framework.sh
+wget https://raw.githubusercontent.com/cp3-llbb/Framework/CMSSW_9_4_X/setup_project_with_framework.sh
+source setup_project_with_framework.sh --branch BRANCH ## select a branch
```
OlivierBondu: maybe we could assume here the default branch should be 94X ? I would expect newcomers (and myself) to copy paste blindly in a terminal without checking that there is need to replace BRANCH by whatever

README.md
```diff
@@ -18,12 +23,14 @@ Common framework for all cp3-llbb analyses
...
 ```
 
-[This script](setup_project_with_framework.sh) will set up a CMSSW release area, apply the recipes in [``bootstrap_jenkins.sh``](bootstrap_jenkins.sh) and [``jenkins_postbuild.sh``](jenkins_postbuild.sh), perform an initial build, and add your and your colleagues' forks on GitHub as remotes for your ``Framework`` clone.
-Through the options `--branch NAME` and `--pr ID`, a project area for a different version can also be created.
+[This script](setup_project_with_framework.sh) will set up a CMSSW release area,
+apply the recipes in [``bootstrap_jenkins.sh``](bootstrap_jenkins.sh) and [``jenkins_postbuild.sh``](jenkins_postbuild.sh),
+perform an initial build, and add your and your colleagues' forks on GitHub as remotes for your ``Framework`` clone.
```
OlivierBondu: speaking of which, maybe we could profit from this cleanup PR to cleanup [these](https://github.com/cp3-llbb/Framework/blob/CMSSW_8_0_6p/first_setup.sh#L17-L27) ? Do you know if there would exist an api to get the list automatically ? This would remove the pain of updating the list...

interface/ElectronsProducer.h
```diff
@@ -52,8 +52,6 @@ class ElectronsProducer: public LeptonsProducer<pat::Electron>, public Identifia
         BRANCH(hcalPFClusterIso, std::vector<float>);
         BRANCH(trackIso, std::vector<float>);
 
-        BRANCH(expectedMissingInnerHits, std::vector<unsigned int>);
-
```
OlivierBondu, line 56: I assume you are removing this because it will be in the ID itself ?

python/ElectronsProducer.py
```diff
@@ -9,34 +9,10 @@
...
             ea_R04 = cms.untracked.FileInPath('cp3_llbb/Framework/data/effAreaElectrons_cone04_pfNeuHadronsAndPhotons.txt'),
             ids = cms.untracked.VInputTag(
-                'egmGsfElectronIDs:cutBasedElectronID-Summer16-80X-V1-veto',
-                'egmGsfElectronIDs:cutBasedElectronID-Summer16-80X-V1-loose',
-                'egmGsfElectronIDs:cutBasedElectronID-Summer16-80X-V1-medium',
-                'egmGsfElectronIDs:cutBasedElectronID-Summer16-80X-V1-tight',
-                'egmGsfElectronIDs:cutBasedElectronHLTPreselection-Summer16-V1',
-                'egmGsfElectronIDs:mvaEleID-Spring16-GeneralPurpose-V1-wp90',
-                'egmGsfElectronIDs:mvaEleID-Spring16-GeneralPurpose-V1-wp80',
-                'egmGsfElectronIDs:mvaEleID-Spring16-HZZ-V1-wpLoose'
```
OlivierBondu, line 19: there is no ID at all ??

bootstrap_jenkins.sh
```diff
@@ -22,15 +22,13 @@ function cms-merge-commit()
...
-cms-merge-commit rafaellopesdesa RegressionCheckNegEnergy 3aafeff0371a1d1eb3db9d95ef50c1a66da25690
-cms-merge-commit cms-met METRecipe_8020_for80Xintegration 92f73cd3d16a9529585865a365de271e0535b68d
+## add recipes here
 
 echo "---> Running git cms-checkdeps"
 git cms-checkdeps -a
 
 echo "---> Adding KaMuCa"
-git clone -o upstream https://github.com/bachtis/analysis.git -b KaMuCa_V4 KaMuCa 
+git clone -o upstream https://github.com/bachtis/analysis.git -b KaMuCa_V4 KaMuCa
```
OlivierBondu: KaMuCa has not been updated in forever, it most probably won't be up to date either for 94X ?

pieterdavid:
I went through the diffs...
For `unit_tests_data.py` the differences are minor: all scale factors gone (I removed them), `electron_expectedMissingInnerHits`, `(fat)jet_pfDeepCSVJetTags_probcc` (it seems not to be there in this version of the deepcsv) and `muon_isMedium2016`. The former and latter will probably be needed still, so I can put them back easily. There are also differences in `jet_pfDeepCSVJetTags_probc` (maybe the merged `probc` and `probcc`?), and in the electron IDs because I removed them... maybe in hindsight I shouldn't have.
For `unit_test_mc_with_db.py` the same, plus the PU weights which I also removed.
For `unit_test_mc.py` there are more differences... I turned off the electron smearing and regression (the recipes changed, see [this twiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/EgammaMiniAODV2#Energy_Scale_and_Smearing), I didn't get to implementing that).
Also the jet, jet systematic variation and MET branches changed. For the former two I suppose that would be because we pick up newer calibrations/recipes somehow, for the MET I'm reasonably certain (I saw code changes that needed the config to be adjusted). Does that make sense?

python/Framework.py
```diff
@@ -69,7 +67,6 @@ def __init__(self, options, verbose=True):
 
         process.options = cms.untracked.PSet(
                 wantSummary = cms.untracked.bool(True),
-                allowUnscheduled = cms.untracked.bool(True)
```
OlivierBondu, line 72: why not keep the option explicit and put it to false instead ?

python/Tools.py
```diff
@@ -185,7 +195,22 @@ def recorrect_met(process, isData, metCollection, jetCollection):
...
+                for vDir in ("Up", "Down"))
+            , [ "pfCandsForUnclusteredUnc", "pfCandsNoJetsNoEleNoMuNoTau", "pfCandsNoJetsNoEleNoMu", "pfCandsNoJetsNoEle", "pfCandsNoJets" ]
+            , [ "patPFMetT1Smear", "patPFMetT1T2SmearCorr", "selectedPatJetsForMetT1T2SmearCorr", "patSmearedJets" ]
+            , ([ "{0}SmearedJetRes{1}".format(mod, vDir) for mod in ("shiftedPat", "shiftedPatMETCorr") for vDir in ("Up", "Down") ] if not isData else
+               [ "patPFMetT1JetRes{0}".format(vDir) for vDir in ("Up", "Down") ])
+            , [ "patPFMetT1SmearJet{0}{1}".format(qty, vDir) for qty in ("Res", "En") for vDir in ("Up", "Down") ]
+            , [ "patPFMetT1Txy", "patPFMetTxyCorr" ]
+        ))+
+        (["ak4PFCHS{0}Corrector".format(corrType) for corrType in ("L1FastL2L3", "L1Fastjet", "L2Relative", "L3Absolute", "L1FastL2L3Residual", "Residual")] if isData else [])+
+        [ "patCaloMet", "patCHSMet", "pfMetCHS", "pfCHS", "patTrkMet", "pfMetTrk", "pfTrk", "metrawCaloNewJEC" ])
```
OlivierBondu, line 213: I will trust you on this one, I don't know JEC nor MET :smile:

test/unit_tests_mc.py
```diff
@@ -11,9 +11,9 @@
 from cp3_llbb.Framework.JetsProducer import discriminators_deepFlavour
 framework.redoJEC(addBtagDiscriminators=discriminators_deepFlavour)
 framework.smearJets()
-framework.applyMuonCorrection("kamuca")
-framework.applyElectronRegression()
-framework.applyElectronSmearing()
+framework.applyMuonCorrection("kamuca", tag="MC_80X_13TeV")
```
OlivierBondu: please comment this line out

OlivierBondu (changes requested):
Thanks!

sorry for the late review, only minor changes :)

python/Tools.py
pieterdavid: This is pure trial&error (but unfortunately needed with tasks, until we get a cms.Task-proof version of it): I went up to the point where I don't get errors about products that don't exist, and neither about modules that are not defined

python/Framework.py
pieterdavid: see https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideUnscheduledExecution#Usage (it's always true, so this does not have any effect)

bootstrap_jenkins.sh
pieterdavid: I only checked it was the latest... disabled now

python/ElectronsProducer.py
pieterdavid: There are 'Fall17' 94X IDs to be used for 2017 and 2018, but the twiki doesn't mention reprocessed 2016 data.

interface/ElectronsProducer.h
pieterdavid: yes, I suppose that with the reprocessing there will be one ID for 2016-2017-2018 that can be used without additional ad-hoc cuts.

README.md
pieterdavid: Of course there is... it return json, so I made the script in python. It takes "active" forks as all those that have been pushed to during the last year, and should also work for other packages - assuming the base repository is set as 'upstream', or under 'cp3-llbb'. There maybe other things missing.

README.md
pieterdavid: Indeed, that makes sense

OlivierBondu (approved):

OlivierBondu:
the `updateremotes` script is just perfect :) thanks for the other fixes !

Do you know why Jenkins fails ? Is it just the outdated ref trees ?

pieterdavid:
I changed the definition of the target branch, so reopening as a new PR, https://github.com/cp3-llbb/Framework/pull/284

PR #281, opened 2018-10-05
Make the tests pass also on SLC7 and with CMSSW_8_0_31, CMSSW_9_3_0 (and above)

pieterdavid:
Two changes to the tests in view of newer releases, see the commit messages for details (the second, about deterministic random generator seeds for the SmearedPatJetProducer, can be removed at some point in the future when CMSSW_8_0_30 is not supported any more).
I will also rebase https://github.com/cp3-llbb/Framework/pull/280 on top of this, these are probably the cause of the unexpected differences there.

### DIFF
--- a/bin/testFrameworkOutput.cc
+++ b/bin/testFrameworkOutput.cc
@@ -14,6 +14,7 @@
 
 #include <TBufferFile.h>
 #include <TLeaf.h>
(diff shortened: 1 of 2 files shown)


alesaggio:
Thanks!

PR #284, opened 2018-11-20
Updates for CMSSW_9_4_X

pieterdavid:
Resubmitting https://github.com/cp3-llbb/Framework/pull/280 since I removed CMSSW_9_4_X and redefined it as the latest CMSSW_8_0_6p commit. Jenkins should only point out differences in the tree content, let's see...


pieterdavid:
OK, the differences in the tests now are mostly missing branches:
   - missing SFs (correct ones to be re-added)
   - electron_expectedMissingInnerHits (additional cut that was applied in 2016, can easily be added again)
   - *_pfDeepCSVJetTags_probcc : it's not there any more, I think probc was defined to be the old probc+probcc
   - muon_isMedium2016
   - no PU weights (1+/-0)
   - different muon PT and E (and relative iso) for MC because I disabled kamuca for the test config
   - different electron momenta (and isolation) for MC because I turned off smearing and regression for now (the number of electrons passing in MC also differs)
   - for electrons in data only the MVA id changed (Run2Fall17IsoV1 now)
   - the other MC (with local JEC db) didn't have electron corrections, so they only have a different MVA id

Please check if I didn't miss anything... I think at this point I can update the reference trees

pieterdavid:
ping... could someone please have a look? This is not meant to provide something usable thing for analysis yet (only a starting point for that), so it does only needs a technical review.
Other question: should we better move this straight to 10_2?

vischia:
Dear all,

I think the PR is fine.
If I don't hear anything before tomorrow midday, I will merge it ;)

vischia:
Merged. I think we don't need to move straight to 10_2, by the way, as 2017 afaik is still in 9_4 (at least as ready-to-use datasets).

Issue #256, opened 2017-09-02
Move test files away from my CERN public folder


blinkseb:
Test files currently used to test PRs are hosted on my public CERN folder. Since it'll soon no longer be accessible, files should be moved away. Files needed to be move are located here:

```
/afs/cern.ch/user/s/sbrochet/www/public/CP3/*.root
```

as well as `/afs/cern.ch/user/s/sbrochet/public/CP3/reduceEdmFile.py` (small CMSSW config file to skim a miniaod file and keep only a bunch of events)

As we discussed IRL a bit yesterday, long-term solution might by that @delaere hosts these files, as he's the most likely to stick around :smile: 

Christophe, can you ping me once you have move the file to a publicly available space (files must be downloadable from outside CERN)?

OlivierBondu:
mmm, the files are now in access denied and it seems there is nothing in `/afs/cern.ch/user/d/delaer/public` and `/afs/cern.ch/user/c/cdelaere/public` is not accessible :cry: 

@blinkseb : any chance you have stored this `reduceEdmFile.py` file somewhere else ?

OlivierBondu:
ha, the files are still in `/afs/cern.ch/user/s/sbrochet/public/CP3` apparently, so it's just a matter of having a cern website set up :smile: 

I copied all the content of that dir to `/afs/cern.ch/user/o/obondu/public/CP3` in the meantime

OlivierBondu:
I put the files in `/eos/user/o/obondu/www/public/CP3` and in `/afs/cern.ch/user/o/obondu/public/CP3` as a temporary solution. The `eos` location is also now tied to https://obondu.web.cern.ch/obondu/public/CP3/ as one can see in the PR https://github.com/cp3-llbb/Framework/pull/259

@delaere : I pretty much followed https://cernbox.web.cern.ch/cernbox/en/web/ to set up the website, so this issue stays open but we have a working solution for the moment it seems :)

pieterdavid:
I'm getting a `503: Service Unavailable.` error... I can push them to my public_html at CP3 for now, but maybe there is a better solution in the mean time?

delaere:
For MoMEMta, we moved files needed by Travis to my public_html... there is no better solution.
See https://github.com/MoMEMta/MoMEMta/blob/master/travis/get-root.sh

We can put them there as well if you like.

pieterdavid:
They should be good in mine for a while as well, and I don't mind... so as you prefer

[...]

PR #283, opened 2018-10-19
Add Tau leptons producer and MET Cov Matrix

clacaputo:
New TauProducer added to the llbb framework: `TauProducer.cc` `TauProducer.h`

Modified:   `METProducer.h` `METProducer.cc`
- MET covariance matrix added

Modified:   `Framework.py`
- new _redoMETsSignificance_ that compute MET covariance and significance
  if JEC or/and JER have been done
- taus producer added to the main sequence

PR #285, opened 2018-12-19
Change test files directory

pieterdavid:
Take the test files from my public_html at CP3, see https://github.com/cp3-llbb/Framework/issues/256#issuecomment-448228307


pieterdavid:
Tests passed with this, so I merged. https://github.com/cp3-llbb/Framework/pull/282 and https://github.com/cp3-llbb/Framework/pull/283 may need to be rebased on top of this (depending on whether Jenkins actually performs the merge when testing).

PR #286, opened 2019-01-08
Added 2017 era

vischia:
Added 2017 era, supposedly.


vischia:
Merge after #287

pieterdavid:
In fact the cleanest would be to rebase... I'm sorry for the mess

vischia:
Mmm Are you sure? From what I see there are a lot of things that differ, no? For example, my PR updates the architecture and CMSSW release, etc, whereas the current (even after #287) branch has still an old architecture and CMSSW release.

pieterdavid:
I see the new ones, e.g. 9_4_9 ([here](https://github.com/cp3-llbb/Framework/blob/CMSSW_9_4_X/CMSSW.release), the default branch is still 8_0_6p), so I am reasonably sure... but if you are not you can always rebase under a new name (since there are only two commits, I would just make a new branch from the lateset `CMSSW_9_4_X` and cherry-pick them)

PR #287, opened 2019-01-08
Updates for CMSSW_9_4_X

pieterdavid:
Same as https://github.com/cp3-llbb/Framework/pull/284 , but now for CMSSW_9_4_X


pieterdavid:
Since it's the same commit hash the code is identical, so the jenkins test are valid, and I think I can merge

PR #288, opened 2019-01-08
Added 2017 era

vischia:
Rebased on top of CMSSW_9_4_X.
Code works in CMSSW_9_4_9 via ttHAnalysis package

### DIFF
--- a/test/TestConfigurationMC.py
+++ b/test/TestConfigurationMC.py
@@ -1,12 +1,13 @@
-
+print('dafuck')
 import FWCore.ParameterSet.Config as cms
 
 from Configuration.StandardSequences.Eras import eras
 from cp3_llbb.Framework import Framework
 from cp3_llbb.Framework.CmdLine import CmdLine
 
-options = CmdLine(defaults=dict(runOnData=0, era="25ns", globalTag='80X_mcRun2_asymptotic_2016_TrancheIV_v7'))
-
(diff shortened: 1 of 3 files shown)


test/TestConfigurationMC.py
```diff
@@ -1,11 +1,11 @@
-
 import FWCore.ParameterSet.Config as cms
 
 from Configuration.StandardSequences.Eras import eras
 from cp3_llbb.Framework import Framework
 from cp3_llbb.Framework.CmdLine import CmdLine
 
-options = CmdLine(defaults=dict(runOnData=0, era="25ns", globalTag='80X_mcRun2_asymptotic_2016_TrancheIV_v7'))
+options = CmdLine(defaults=dict(runOnData=0, era="25ns", globalTag='94X_mc2017_realistic_v14'))
```
pieterdavid: It would be cleaner to change to `era="2017"` as well then (but they weren't consistent before - and I'm not sure this test config is even working still, those used by the test are called `unit_test_*.py`

pieterdavid (commented):
Only one minor comment, otherwise it all looks fine to me

test/TestConfigurationMC.py
vischia: True, at that stage I had not yet seen the era thing. Will update the PR

vischia:
Done (also for the TestConfigurationData.py, and I have updated the paths to the test files, since I was at it, taking them from the ones I had tested in https://github.com/vischia/ttHAnalysis/blob/master/test/ttHConfiguration.py

PR #289, opened 2019-01-17
[WIP] First series of updates for 10_2_X

pieterdavid:
Summary:
   - this makes the framework run with CMSSW 10_2_9, which is recommended for the combined 2016-2017-2018 dataset (rereco up to September 2018, then prompt; re-MiniAOD for 2016)
   - updated electron IDs (Fall17 V2)
   - updated to the latest POG recipes electron scale/smearing and JEC/JER (removed quite some of the 94X `cms.Task` hacks)
   - (technical) update for setup script (and needed to add the requests python package for `updateremotes`), silence GCC7 warning about non-virtual destructors for virtual base classes
   - changed the default `SCRAM_ARCH` to CentOS7, since much of the computing resources is migrating or will do so soon (e.g. lxbatch, grid sites, the cp3 partition on slurm)

Still to do:
   - [ ] verify that the electron corrections are actually being applied
   - [ ] add 2017 and 2018 data and MC to the tests (era, GT, input files and specifiic configs/recipes, if needed)
   - [ ] re-enable fat jets (user float names etc. changed between PUPPI and AK8PF, so I took out the producer for now to save time)
   - [ ] update reference trees when we are happy about all the above
   - [ ] configure Jenkins to test with CentOS7 as well, at some point
   - [ ] backport https://github.com/cp3-llbb/Framework/pull/282


vischia:
The "details" say "not found", so I guess the test has been run so many days ago that the outcome has been cleaned up by the system. Is there a way of forcing again the test?

pieterdavid:
Indeed. I think it's possible to retrigger the build in jenkins, but this one is not ready for tests yet anyway (and the test configs will need to change if we want some coverage for the "full run2" sample).

